PR #4, opened 2020-02-26
Add license headers to our scripts

iliana:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket/issues/469

**Description of changes:**

We normally don't add license headers to files in Bottlerocket, but `sheltie` and `start_admin_sshd.sh` are the only files in the admin container written by us (in this case, the Bottlerocket developers) that warrant license information. In this case, we believe a copyright header is the right thing instead of adding COPYRIGHT and LICENSE-*.

**Testing done:**

`make` works.

**Terms of contribution:**

### DIFF
--- a/sheltie
+++ b/sheltie
@@ -1,4 +1,7 @@
 #!/bin/bash
+# This file is part of Bottlerocket.
+# Copyright Amazon.com, Inc., its affiliates, or other contributors. All Rights Reserved.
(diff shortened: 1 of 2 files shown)


tjkirch (approved):
(technical approval)

zmrow (approved):
📌

ajorg-aws (approved):

PR #10, opened 2020-06-01
Use empty anonymous volumes to test with

jahkeup:
**Issue number:**

n/a

**Description of changes:**

Use newer syntax to specify an anonymous volume to mount and mask libraries. Without this patch some versions of Docker may non-deterministically copy over compatible libraries and render the test ineffective.

**Testing done:**

`make`

**Terms of contribution:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ dist: all
 
 # Build the container image.
 build:
-	docker build \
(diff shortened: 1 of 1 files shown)


samuelkarp (approved):

zmrow (approved):
🍚

PR #14, opened 2020-08-13
Use buildkit for docker builds

rothgar:
**Description of changes:**
Switch from old docker build environment to docker buildkit by default

**Testing done:**
Built the container a few times with and without buildkit. Builds were about 30s faster

Buildkit
real    2m12.879s
user    0m1.077s
sys     0m0.536s

without Buildkit
real    2m45.331s
user    0m0.400s
sys     0m0.294s

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ dist: all
 	docker save $(IMAGE_NAME) | gzip > $(DISTFILE)
 
 # Build the container image.
+build: export DOCKER_BUILDKIT = 1
 build:
 	docker build $(DOCKER_BUILD_FLAGS) \
 		--tag $(IMAGE_NAME) \


Makefile
```diff
@@ -35,6 +35,7 @@ dist: all
 	docker save $(IMAGE_NAME) | gzip > $(DISTFILE)
 
 # Build the container image.
+build: export DOCKER_BUILDKIT = 1
 build:
 	docker build $(DOCKER_BUILD_FLAGS) \
```
samuelkarp, line 40:
```suggestion
build:
	DOCKER_BUILDKIT=1 docker build $(DOCKER_BUILD_FLAGS) \
```

This looks a bit cleaner to me since we don't need to export `DOCKER_BUILDKIT` for all of the commands in the `build` target (not that there are any others right now...).  Not a strong preference though.

samuelkarp (approved):

Makefile
rothgar: I'm fine with either. The only reason I typically add to all commands is if you end up with multiple docker commands in the build target it can export to all of them at once instead of adding to each.

webern (approved):

tjkirch:
Hey @rothgar, I wanted to apologize for this PR getting lost and not merged after approval.  Thanks again for your contribution!

Issue #17, opened 2021-02-03
Support using keys from IMDS as TrustedUserCAKeys


backjo:
**Image I'm using:**
Latest (v0.5.2)

**Issue or Feature Request:**
Feature Request!

Today, we're using a custom BottleRocket admin container to achieve certificate-based authentication for SSH users - this issue is mainly to determine whether this change would be welcome upstream. 

We're adding the following line to our sshd_config:
```
TrustedUserCAKeys /home/ec2-user/.ssh/authorized_keys
```

This allows us to upload the public part of our signing key(s) into AWS, and then manage SSH access through issuance of signed certificates. I think the main downside into enabling this is that there might be folks who for some reason are using their SSH private keys today to also do signing with an 'ec2-user' principal - though that seems fairly unlikely.

tjkirch:
Thanks for the suggestion!  @jahkeup said he's used this as well, and had to specify both `TrustedUserCAKeys` and `AuthorizedKeysFile` to make sure both options are still effective.  I'm curious whether you can point them at the same file, and the format that matches will take effect?

backjo:
@tjkirch it turns out you can! I tested out the changes in #18 , and was able to authenticate with an authorized key, as well as a CA-signed key

jpculp:
@backjo, I just wanted to follow-up that we included the ability to specify `TrustedUserCAKeys` with the change to the admin container logic to use public keys from user-data (#19). This feature is available as of [v0.6.0](https://github.com/bottlerocket-os/bottlerocket-admin-container/releases/tag/v0.6.0) and included in [Bottlerocket v1.0.6](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.0.6)! I encourage you to check out the example in the [authenticating with the admin container](https://github.com/bottlerocket-os/bottlerocket-admin-container/#authenticating-with-the-admin-container) documentation. Thank you again for the suggestion!

backjo:
hey @jpculp  - thanks for pinging! We've actually been on 1.0.7 and using this for a while - works great!

PR #18, opened 2021-02-05
Add authorized_keys to TrustedUserCAKeys for certificate-based authen…

backjo:
…tication

**Issue number:**
#17 

**Description of changes:**
Allows the authorized_keys file to also be used for certificate-based authentication. For larger organizations, certificate based authentication is desirable because it allows for short-lived credentials and allows for the CA to be the decider of who has access to which machines. 

**Testing done:**
- Tested ssh access on EC2 with both a public/private pair (existing use case) as well as with a different public/private pair signed by a CA (with the CA's public key being the one in authorized_keys)

**Terms of contribution:**

### DIFF
--- a/sshd_config
+++ b/sshd_config
@@ -3,7 +3,7 @@ HostKey /.bottlerocket/host-containers/admin/etc/ssh/ssh_host_ecdsa_key
 HostKey /.bottlerocket/host-containers/admin/etc/ssh/ssh_host_ed25519_key
 
 PasswordAuthentication no
-
+TrustedUserCAKeys /home/ec2-user/.ssh/authorized_keys
 ChallengeResponseAuthentication no
 
 UsePAM yes
--- a/sshd_config
+++ b/sshd_config
@@ -3,8 +3,11 @@ HostKey /.bottlerocket/host-containers/admin/etc/ssh/ssh_host_ecdsa_key
 HostKey /.bottlerocket/host-containers/admin/etc/ssh/ssh_host_ed25519_key
 
 PasswordAuthentication no
-TrustedUserCAKeys /home/ec2-user/.ssh/authorized_keys
 ChallengeResponseAuthentication no
+
+# Allow authentication with standard key pairs or certificates.
+AuthorizedKeysFile .ssh/authorized_keys
+TrustedUserCAKeys /home/ec2-user/.ssh/authorized_keys
 
 UsePAM yes
 


sshd_config
```diff
@@ -3,7 +3,7 @@ HostKey /.bottlerocket/host-containers/admin/etc/ssh/ssh_host_ecdsa_key
 HostKey /.bottlerocket/host-containers/admin/etc/ssh/ssh_host_ed25519_key
 
 PasswordAuthentication no
-
+TrustedUserCAKeys /home/ec2-user/.ssh/authorized_keys
```
tjkirch: With `AuthorizedKeysFile`, this can be a path relative to the home directory of the user logging in, and that's the form used by default (`.ssh/authorized_keys`).  The man page isn't clear, but I suspect `TrustedUserCAKeys` should work the same way, and then we wouldn't have to hardcode the username.  Do you know if that works here?  On other platforms the user might not be ec2-user, so it'd be nice to avoid specifying when possible, easing the clean-up later.

I also think it'd be a good idea to explicitly list `AuthorizedKeysFile`, even though it's the default, to make it clear that we expect to accept both.

So, something like this, if it works:  (not sure on the best comment wording)
```
PasswordAuthentication no
ChallengeResponseAuthentication no

# Allow authentication with standard key pairs or certificates.
AuthorizedKeysFile .ssh/authorized_keys
TrustedUserCAKeys .ssh/authorized_keys
```

tjkirch (approved):
Thank you!

sshd_config
backjo: @tjkirch tested this out locally - AuthorizedKeysFile does work as a relative path as the docs indicate, but TrustedUserCAKeys does not use the path as a relative path. Verified this by first testing with `TrustedUserCAKeys .ssh/authorized_keys`, which did not work. Copying the file in the container to `/.ssh/authorized_keys` allowed me to login with my signed certificate - and then removing that copied file again denied me access.

jahkeup:
Always cool to see SSH-CA's being put to use @backjo 😺 

That said, I'm not sure that this belongs in the base container image that we have today and that is used by default on Bottlerocket. 

The `TrustedUserCAKeys` directive may induce logged warning and authentication "error" messages that  don't apply to the typical use case - where its good 'ol `.ssh/authorized_keys` being used. These extra messages may confuse user and/or mislead monitoring infrastructure. With this being a more advanced, atypical configuration (taking advantage of IMDS carrying CA keys in - again, very cool!) I think that a custom image (for you) would be cleaner and clearer for your use-case.

Without a mechanism to optionally enable this feature, I'm hesitant to merge and enable this by default and for all instances of the admin host container.

backjo:
Yeah - we're already using a custom image to achieve this, so it doesn't need to be merged and enabled to solve for us and our use case (though it  would be nice to be on upstream :) ). 

For the user experience, I don't think adding this would add too much confusion. The error messages (in my experience testing, at least) don't change just because of the presence of this setting if users are attempting to login with standard public/private keypairs. The only time that the user would get a new message was if they attempted to sign in with a certificate that was invalid (either expired, an invalid signer, or invalid principal, primarily) - in which case, I think it'd be acceptable to assume that the user is aware of SSH certificates and that the error would more useful than confusing. 

I'd advocate that with the heavy focus on Security features in BottleRocket today (which is why we're leaning so heavily into it), having this feature would outweigh user confusion - at this point, mostly because it's unclear that any additional confusion would be caused for basic SSH users. 

It also sounds like creating a mechanism to pass flags into the admin container (maybe passing TOML settings under a certain prefix as env variables?) might be something to open an issue for.

tjkirch:
Hey @backjo!  I wanted to follow up here because your use case is important to us, and I didn't want you to think we were just letting it sit.

We're currently working on some improvements to the admin container, and the interface between the container and Bottlerocket, to make things like this more sustainable.  Specifically, we're moving some AWS-specific logic into Bottlerocket, so it doesn't have to be repeated by alternate host containers and to make supporting other platforms easier as well.

The [admin container PR](https://github.com/bottlerocket-os/bottlerocket-admin-container/pull/19) in particular handles adding TrustedUserCAKeys to sshd_config if the user specified a value in user data.  (Similar to the flags you mentioned.)  It's still undergoing some changes, but I'm hoping it will address your request here, and I'd be interested to get your thoughts on the approach.

backjo:
Hey @tjkirch  - appreciate the update! The PR looks good - UserData makes a lot of sense! Only thing I would be ever so slightly worried about is the impact of ssh keys there to the overall size of the user data. But directionally, it totally seems to be something we'd use and get value out of :)

tjkirch:
@backjo Thanks!  Yeah, user data is a limited resource, so we don't intend to store much there, but configuration that's important early in the startup process can be a good fit.  (I'm working on https://github.com/bottlerocket-os/bottlerocket/issues/1320 to allow compression and extend it a bit.)

PR #19, opened 2021-02-18
Change admin container logic to use public keys from user-data

jpculp:
**Issue number:**

N/A

**Description of changes:**

Adds jq to the container image.

Replaces IMDS logic with fetching public keys from user-data.

This change allows the admin container to be platform agnostic as it
retrieves its .ssh/authorized_keys from a base64-encoded 
JSON-structured block in user-data rather than relying on the AWS
instance metadata service (IMDS) directly.

Example JSON: `{"ssh":{"authorized_keys":[]}}`

Also adds support for TrustedUserCAKeys.

This allows users who want to set TrustedUserCAKeys to do so by adding
them to the base64-encoded JSON-structured user-data block.

Example JSON: `{"ssh":{"trusted_user_ca_keys":[]}}`

Bumps version to v0.6.0.

**Testing done:**
Launched the new admin container on a bottlerocket instance containing [shibaken](https://github.com/bottlerocket-os/bottlerocket/pull/1331) and verified that SSH access was still available.

After that, I went through several test cases of disabling the admin container, posting a new base64-encoded block to `host-containers.admin.user-data`,  re-enabling the admin container, and checking the connection.
- :white_check_mark: Empty JSON Structure (expected fail)
- :white_check_mark: Custom authorized key
- :white_check_mark: Several keys, some good, some bad (only the strings that could be parsed by ssh-keygen were present in `authorized_keys`)
- :white_check_mark: Authorized key AND trusted user CA key (`.ssh/authorized_keys`, `/etc/ssh/trusted_user_ca_keys.pub`, and `/etc/ssh/sshd_config` looked as expected)
- :white_check_mark: Removed authorized key so that only trusted user CA key was available

**Terms of contribution:**


start_admin_sshd.sh
```diff
@@ -8,46 +8,86 @@ mkdir -p /home/ec2-user/.ssh/
...
   fi
   echo "${public_key_data}" >> "${ssh_authorized_keys}"
   if ! grep -q "${public_key_data}" "${ssh_authorized_keys}"; then
-    echo "Failed to write key with index ${public_key_index} to authorized_keys" >&2
+    echo "Failed to write key ${public_key_data} to ${ssh_authorized_keys}" >&2
     continue
   fi
 done
 
 # If we didn't write any keys at all, there's not much point in continuing
```
zmrow: I don't think we want to exit here as there's a possibility the user is only using CA keys and not SSH keys.

README.md
```diff
@@ -11,3 +11,33 @@ For more information about how the admin container fits into the Bottlerocket op
...
 Then run `make`!
+
+## Authenticating with the Admin Container
+
+Starting from v0.6.0, users have the option to pass in their own ssh keys rather than the admin container relying on the AWS instance metadata service (IMDS).
+
+Users can add their own keys by populating the admin container's user-data with a base64-encoded JSON block.
+(If you would like to continue using the public keys available in IMDS this is already done for you!)
+
+To use custom public keys to .ssh/authorized_keys you will want to generate a JSON-structure like this:
```
jpculp: This is awkward. I'll fix this after I get input for the rest of the README.

README.md
```diff
@@ -11,3 +11,33 @@ For more information about how the admin container fits into the Bottlerocket op
...
+          hz2ItxCih+PnDSUaw+WNQn/mZphTk/a/gU8jEzoOWbkM4yxyb/wB96xbiFveSFJuOp/d6RJhJOI0iBXr
+          lsLnBItntckiJ7FbtxJMXLvvwJryDUilBMTjYtwB+QhYXUMOzce5Pjz5/i8SeJtjnV3iAoG/cQk+0FzZ
+          qaeJAAHco+CY/5WrUBkrHmFJr6HcXkvJdWPkYQS3xqC0+FmUZofz221CBt5IMucxXPkX4rWi+z7wB3Rb
+          BQoQzd8v7yeb7OzlPnWOyN0qFU0XA246RA8QFYiCNYwI3f05p6KLxEXAMPLE my-key-pair"
+      ]
+   }
+}
+```
+
+Once you've created your JSON, you'll need to base64-encode it and set the value to the user-data in your [instance user data toml](https://github.com/bottlerocket-os/bottlerocket#using-user-data).
```
zmrow: `...set it as the value of the admin host container's user-data setting...`

README.md
```diff
@@ -11,3 +11,33 @@ For more information about how the admin container fits into the Bottlerocket op
 
 You'll need Docker 17.06.2 or later, for multi-stage build support.
 Then run `make`!
+
+## Authenticating with the Admin Container
+
+Starting from v0.6.0, users have the option to pass in their own ssh keys rather than the admin container relying on the AWS instance metadata service (IMDS).
+
+Users can add their own keys by populating the admin container's user-data with a base64-encoded JSON block.
```
zmrow, line 19: It's probably worth mentioning that if the user-data setting is populated for this container, we aren't going to fetch any keys.

README.md
```diff
@@ -11,3 +11,33 @@ For more information about how the admin container fits into the Bottlerocket op
...
+(If you would like to continue using the public keys available in IMDS this is already done for you!)
+
+To use custom public keys to .ssh/authorized_keys you will want to generate a JSON-structure like this:
+
+```
+{
```
zmrow, line 25: Let's add an example of the entire supported JSON structure for user reference.

jahkeup:
Related to changes in PR bottlerocket-os/bottlerocket#1331

README.md
jahkeup: > It's probably worth mentioning that if the user-data setting is populated for this container, we aren't going to fetch any keys.

Not just not fetch any keys, but also not handle populating settings for **anything else** at this time - even if there aren't other pieces of data yet. If this is the expected handling (for the foreseeable future) then we should help users understand that they may be opting out of *any* populated settings (at least until that changes..).

README.md
```diff
@@ -11,3 +11,33 @@ For more information about how the admin container fits into the Bottlerocket op
...
+   "ssh":{
+      "authorized_keys":[
+         "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQClKsfkNkuSevGj3eYhCe53pcjqP3maAhDFcvBS7O6V
+          hz2ItxCih+PnDSUaw+WNQn/mZphTk/a/gU8jEzoOWbkM4yxyb/wB96xbiFveSFJuOp/d6RJhJOI0iBXr
+          lsLnBItntckiJ7FbtxJMXLvvwJryDUilBMTjYtwB+QhYXUMOzce5Pjz5/i8SeJtjnV3iAoG/cQk+0FzZ
+          qaeJAAHco+CY/5WrUBkrHmFJr6HcXkvJdWPkYQS3xqC0+FmUZofz221CBt5IMucxXPkX4rWi+z7wB3Rb
+          BQoQzd8v7yeb7OzlPnWOyN0qFU0XA246RA8QFYiCNYwI3f05p6KLxEXAMPLE my-key-pair"
```
jahkeup: I think the embedded newlines may confuse some users - can we show this as it "should" be provided? As a single line?

README.md
```diff
@@ -11,3 +11,33 @@ For more information about how the admin container fits into the Bottlerocket op
...
+      ]
+   }
+}
+```
+
+Once you've created your JSON, you'll need to base64-encode it and set the value to the user-data in your [instance user data toml](https://github.com/bottlerocket-os/bottlerocket#using-user-data).
+
+```
+[settings.host-containers.admin]
+user-data = "eyJzc2giOnsiYXV0aG9yaXplZF9rZXlzIjpbXX19"
```
jahkeup: This example would be helped by showing the text (and process, as shown here) that produces the inline base64 value:

```suggestion
[settings.host-containers.admin]
# ex: echo '{"ssh":{"authorized_keys":[]}}' | base64
user-data = "eyJzc2giOnsiYXV0aG9yaXplZF9rZXlzIjpbXX19"
```

start_admin_sshd.sh
```diff
@@ -8,36 +8,86 @@ mkdir -p /home/ec2-user/.ssh/
 chmod 700 /home/ec2-user/.ssh/
 ssh_host_key_dir="/.bottlerocket/host-containers/admin/etc/ssh"
 ssh_config_dir="/home/ec2-user/.ssh"
+user_data="/.bottlerocket/host-containers/admin/user-data"
```
jahkeup: nit: this, and the above, are constants - they should be declared read only:
 
```suggestion
declare -r user_data="/.bottlerocket/host-containers/admin/user-data"
```

start_admin_sshd.sh
```diff
@@ -8,36 +8,87 @@ mkdir -p /home/ec2-user/.ssh/
...
-    echo "Key ${public_key_data} with index ${public_key_index} looks invalid" >&2
-    continue
+set -f
+IFS='
+'
+
+# Is user-data does not contain a JSON block, there's not much point in continuing.
+if ! jq empty ${user_data} &>/dev/null; then
+    echo "${user_data} does not contain valid JSON." >&2
+fi
```
jahkeup: I'm not sure this error message accurately describes the situation. There's more than one cases covered here:

- `test -s "$user_data"` - the file is empty
- `jq -e 'empty' < "$user_data" - the provided data was empty (any of `<nothing>`, `[]`)

I think we actually would want to check, either in addition or in place of some of the above:

- `jq -e 'type == "object"' < "$user_data"` - verify that we have an object to read into (in the if: `jq '(type != "object) and halt_error' < "$user_data"`)


No matter what, the `jq` command should be using `-e` to set the exit code based on the result.

---

This comment seems off based on the nearby code. Were you planning on exiting here?

```suggestion
# Is user-data does not contain a JSON block, there's not much point in continuing.
if ! jq empty ${user_data} &>/dev/null; then
    echo "${user_data} does not contain valid JSON." >&2
    exit $no_keys_exit_status
fi
```

start_admin_sshd.sh
```diff
@@ -8,36 +8,87 @@ mkdir -p /home/ec2-user/.ssh/
...
+
+# Populate authorized_keys with all the public keys found in api user-data
+if grep -q authorized_keys ${user_data}; then
+  ssh_authorized_keys="${ssh_config_dir}/authorized_keys"
+  touch ${ssh_authorized_keys}
+  chmod 600 ${ssh_authorized_keys}
```
jahkeup: There's mixed use of `${varname}`, `"${varname}"` and `"$varname"` - can you update this to use one of the quoted variants? I'm partial to `"$varname"` where possible, but am okay with `"${varname}"` too.

```suggestion
if grep -q authorized_keys "$user_data"; then
  ssh_authorized_keys="$ssh_config_dir/authorized_keys"
  touch "$ssh_authorized_keys"
  chmod 600 "$ssh_authorized_keys"
```

start_admin_sshd.sh
```diff
@@ -8,36 +8,87 @@ mkdir -p /home/ec2-user/.ssh/
...
+  # If we didn't write any keys at all, then this isn't a valid auth method
+  if [ ! -s "${ssh_trusted_user_ca_keys}" ]; then
+    echo "Failed to write any valid public keys to ${ssh_trusted_user_ca_keys}" >&2
+  else
+    printf "\nTrustedUserCAKeys: %s\n" "${ssh_trusted_user_ca_keys}" >> ${sshd_config}
+    # If we didn't append the TrustedUserCAKeys, then this isn't a valid auth method
+    if ! grep -q "^TrustedUserCAKeys: ${ssh_trusted_user_ca_keys}$" ${sshd_config}; then
+      echo "Failed to write TrustedUserCAKeys to sshd_config" >&2
+    else
+      ((++available_auth_methods))
```
jahkeup: Typically `++n` is used in very specific use cases and tickles my spideysenses - can we switch to `n++`?

```suggestion
      ((available_auth_methods++))
```

start_admin_sshd.sh
```diff
@@ -8,36 +8,87 @@ mkdir -p /home/ec2-user/.ssh/
...
-    | xargs))
-
-for public_key_index in "${public_key_indexes[@]}"; do
-  public_key_data="$(curl -H "${imds_request_add_header}" -sf "${public_key_base_url}/${public_key_index}/openssh-key")"
-  if [[ ! "${public_key_data}" =~ ^"ssh" ]]; then
-    echo "Key ${public_key_data} with index ${public_key_index} looks invalid" >&2
-    continue
+set -f
+IFS='
+'
```
jahkeup: Please use the escaped variant:

```suggestion
IFS=$'\n'
```

start_admin_sshd.sh
```diff
@@ -8,36 +8,87 @@ mkdir -p /home/ec2-user/.ssh/
...
-public_key_indexes=($(curl -H "${imds_request_add_header}" -sf "${public_key_base_url}" \
-    | cut -d= -f1 \
-    | xargs))
-
-for public_key_index in "${public_key_indexes[@]}"; do
-  public_key_data="$(curl -H "${imds_request_add_header}" -sf "${public_key_base_url}/${public_key_index}/openssh-key")"
-  if [[ ! "${public_key_data}" =~ ^"ssh" ]]; then
-    echo "Key ${public_key_data} with index ${public_key_index} looks invalid" >&2
-    continue
+set -f
```
jahkeup: I'm curious why `set -f` is introduced. Was there an edge case that was being bumped up against?

start_admin_sshd.sh
```diff
@@ -8,36 +8,87 @@ mkdir -p /home/ec2-user/.ssh/
...
+      echo "Failed to write TrustedUserCAKeys to sshd_config" >&2
+    else
+      ((++available_auth_methods))
+    fi
   fi
-done
+fi
+
+unset IFS
+set +f
```
jahkeup: 😢 

Can we save the `OLDIFS` before setting `IFS` and then restore it here instead?
```suggestion
IFS="$OLDIFS"
set +f # (setf ... 'lol)
```

jahkeup (changes requested):
Small scripty changes - thanks for rolling this up with bottlerocket-os/bottlerocket#1331 👍🏽

start_admin_sshd.sh
```diff
@@ -8,36 +8,87 @@ mkdir -p /home/ec2-user/.ssh/
...
 
-# If we didn't write any keys at all, there's not much point in continuing
-if [ ! -s "${ssh_authorized_keys}" ]; then
-  echo "Failed to write any valid public keys to authorized_keys" >&2
+# If none of the above was sucessful, then users cannot authenticate
```
bcressey: nit: "successful"

README.md
```diff
@@ -11,3 +11,40 @@ For more information about how the admin container fits into the Bottlerocket op
 
 You'll need Docker 17.06.2 or later, for multi-stage build support.
 Then run `make`!
+
+## Authenticating with the Admin Container
+
+Starting from v0.6.0, users have the option to pass in their own ssh keys rather than the admin container relying on the AWS instance metadata service (IMDS).
+
+Users can add their own keys by populating the admin container's user-data with a base64-encoded JSON block.
+If user-data is populated then bottlerocket will not fetch from IMDS at all, but if user-data is not set then bottlerocket will continue to use the keys from IMDS.
```
bcressey:
```suggestion
If user-data is populated then Bottlerocket will not fetch from IMDS at all, but if user-data is not set then Bottlerocket will continue to use the keys from IMDS.
```

README.md
```diff
@@ -11,3 +11,40 @@ For more information about how the admin container fits into the Bottlerocket op
...
+
+To use custom public keys for .ssh/authorized_keys and/or custom CA keys for .ssh/trusted you will want to generate a JSON-structure like this:
```
bcressey:
```suggestion
To use custom public keys for `.ssh/authorized_keys` and/or custom CA keys for `.ssh/trusted` you will want to generate a JSON-structure like this:
```

start_admin_sshd.sh
jpculp: Probably shouldn't be there actually.

start_admin_sshd.sh
jpculp: I gave this a shot, but the container seemed to stop immediately after the first instance:
```
+ [[ ! -s /home/ec2-user/.ssh/authorized_keys ]]
+ (( available_auth_methods++ ))
```
I think this is due to the `set -e` at the top of the script.

zmrow (approved):
I think this looks good - can you update the testing section in the PR description with the testing you've done?

:volleyball:

etungsten (approved):

start_admin_sshd.sh
jahkeup: Ah, yeah this would have to written as `: (( available_auth_methods++ ))` or `let available_auth_methods++`.

start_admin_sshd.sh
```diff
@@ -6,38 +6,87 @@ set -e
...
+      echo "Failed to write key ${trusted_key_data} to ${ssh_trusted_user_ca_keys}" >&2
+      continue
+    fi
+  done
+
+  # If we didn't write any keys at all, then this isn't a valid auth method
+  if [[ ! -s "${ssh_trusted_user_ca_keys}" ]]; then
+    echo "Failed to write any valid public keys to ${ssh_trusted_user_ca_keys}" >&2
+  else
+    printf "\nTrustedUserCAKeys %s\n" "${ssh_trusted_user_ca_keys}" >> "${sshd_config}"
```
jahkeup: nit: printf is a bit muddier than the equivalent `echo`:

```suggestion
    echo "TrustedUserCAKeys ${ssh_trusted_user_ca_keys}" >> "${sshd_config}"
```

```bash
# or if the extra newline is needed:
echo -e "\nTrustedUserCAKeys ${ssh_trusted_user_ca_keys}" >> "${sshd_config}"
```

start_admin_sshd.sh
```diff
@@ -6,38 +6,87 @@ set -e
...
+  touch "${ssh_authorized_keys}"
+  chmod 600 "${ssh_authorized_keys}"
+
+  userdata_authorized_keys=($(jq -r '.ssh.authorized_keys[]' "${user_data}" 2> /dev/null))
+
+  for public_key_data in "${userdata_authorized_keys[@]}"; do
+    if [[ ! "${public_key_data}" =~ ^"ssh" ]]; then
+      echo "Key ${public_key_data} looks invalid" >&2
+      continue
+    fi
```
jahkeup: The admin container has the ssh tools in it, I think using them is a safer bet than using a regex to validate the key. If there's changes later on, then we get that handling updated with the packaging updates.

```bash
ssh-keygen -e -f .ssh/id_rsa.pub # check a file
ssh-keygen -e -f <(echo "$public_key_data") # check the string
```

```bash
+ ssh-keygen -e -f .ssh/id_rsa.pub
---- BEGIN SSH2 PUBLIC KEY ----
Comment: "2048-bit RSA, converted by jakeev@"
AAAAB3NzaC1yc2EAAAADAQABAAABAQDjewtGdLslHmWR1YJVgQjKoSxd5JggOgzV6De+Tj
QXe0/KRxsomdL7EYdE8v6urG4bzsH8/7laSEqKvLzdZQI/yktF5gdBrDmvoD44O0D/QNPe
xeIkn+36+p/ruzyzrDPzKDp4w9B4Ze2ltuOtkiaSypBx1pYNSWBfLQBrHck9XGVfGQDu32
bDs91ViTBkvHqkfxP9fgV1CMUP7xIFy7qeI+MzBxUivCd7XDwiLPmwoSYTf+LEttA5BhPo
eECI0tVsJoog80XpS/C7ThSMMyNoAwKLVtiZK223CnYIIIrw07D1+2wWIFVD9uh9ImWup2
iEJ/V3y07qx3GC+05mP53Z
---- END SSH2 PUBLIC KEY ----
+ echo 0
0
+ ssh-keygen -e -f /dev/null
Load key "/dev/null": invalid format
+ echo 255
255
```

start_admin_sshd.sh
jpculp: It seems those also are causing an early exit.

start_admin_sshd.sh
```diff
@@ -6,38 +6,87 @@ set -e
...
+OLDIFS="${IFS}"
+IFS=$'\n'
+
+# Is user-data does not contain a JSON block, there's not much point in continuing.
+if [[ ! $(jq empty "${user_data}" &>/dev/null) ]]; then
+  echo "${user_data} does not contain valid JSON." >&2
+fi
+
+# Populate authorized_keys with all the public keys found in api user-data
+if grep -q authorized_keys "${user_data}"; then
```
jahkeup: I'm not sure that the "invalid json check" will trigger, unless the file doesn't exist:

```bash
jq empty /dev/null ; echo $?
# prints: 0
```

This might be due to a difference in the version of jq used, but I'm thinking that it might be irrelevant. Instead, the follow up check can check for this.

```bash
if ! jq -e '.["ssh"]["authorized_keys"] as $val | ($val | length > 0)' <<< "$json_data"
then
  log "ERROR: user data does not specify ssh.authorized_keys"
else
  # handle ssh-authorized keys
fi
```
<details>

```
( for key in value_key array_key empty_key empty_array null_key; do ( set -x; jq -e --arg key $key '.[$key]? as $val | ($val | length > 0)' <<< '{"value_key": "imavalue", "array_key": [42], "empty_key": "", "empty_array": [], "null_key": null}' ; ); echo $?; done )
+ jq -e --arg key value_key '.[$key]? as $val | ($val | length > 0)'
true
0
+ jq -e --arg key array_key '.[$key]? as $val | ($val | length > 0)'
true
0
+ jq -e --arg key empty_key '.[$key]? as $val | ($val | length > 0)'
false
1
+ jq -e --arg key empty_array '.[$key]? as $val | ($val | length > 0)'
false
1
+ jq -e --arg key null_key '.[$key]? as $val | ($val | length > 0)'
false
1
```

</details>

In addition, lifting the procedure into a function will allow a cleaner short circuiting.

```bash
process_ssh_authorized_keys() {
    # allow injected values for testing
	local json_input="${1:-/dev/stdin}"
    if ! jq -e '.["ssh"]["authorized_keys"] as $val | ($val | length > 0)' < "$json_input"
    then
        log "ERROR: user data does not specify ssh.authorized_keys"
        return 1
    fi

    # handle ssh-authorized keys
    
    # print collected keys for installation
    ( IFS=$'\n'; echo "${validated_keys[*]}" )
}

# ... in script ...
if authorized_keys=$(process_authorized_keys); then
    install -D /dev/null -m 0600 -o ec2-user "$ec2_user/.ssh/authorized_keys"
    echo "$authorized_keys" > "$ec2_user/.ssh/authorized_keys"
    let auth_available++
fi
```

start_admin_sshd.sh
jahkeup: > It seems those also are causing an early exit.

TIL:

> from `bash -c 'help let'`
>
>     Exit Status:
>     If the last ARG evaluates to 0, let returns 1; let returns 0 otherwise.

The `: ((available_auth_methods++))` expression should be okay no matter what (because of `:` noop).

<details>

```bash
( x=0; set -x; : x++; echo $?; x=0; let x++; echo $?; x=0; let ++x; echo $?)
# + : x++
# + echo 0
# 0
# + x=0
# + let x++
# + echo 1
# 1
# + x=0
# + let ++x
# + echo 0
# 0
```

</details>

start_admin_sshd.sh
jpculp: jq empty seemed to work in my testing, but I'll admit it's always felt a little hacky. This seems more robust.

start_admin_sshd.sh
jahkeup: Or:

```bash
cat >> "${sshd_config}" <<EOF
# Configured by user data
TrustedUserCAKeys $ssh_trusted_user_ca_keys
EOF
```

start_admin_sshd.sh
jpculp: I decided to change the wording here, so no more typo to fix.

etungsten (approved):

start_admin_sshd.sh
```diff
@@ -6,38 +6,94 @@ set -e
...
+}
+
+# Populate trusted_user_ca_keys with all the authorized keys found in user-data
+if trusted_user_ca_keys=$(parse_trusted_user_ca_keys); then
+  ssh_trusted_user_ca_keys="/etc/ssh/trusted_user_ca_keys.pub"
+  sshd_config="/etc/ssh/sshd_config"
+  install -D -m 0600 /dev/null "${ssh_trusted_user_ca_keys}"
+  echo "${trusted_user_ca_keys}" > "${ssh_trusted_user_ca_keys}"
+
+  cat <<EOF >> "${sshd_config}"
```
jahkeup: Please put the heredoc starting token at the end of the redirections:

```suggestion
  cat >> "${sshd_config}" <<EOF
```

start_admin_sshd.sh
```diff
@@ -6,38 +6,94 @@ set -e
...
+
```
jahkeup: This extra newline seems unneeded - the comment adds context if the intent is to break up the appended config from the prior. This isn't a sticking point though, if you want to keep this: feel free to leave it.

zmrow (approved):

start_admin_sshd.sh
```diff
@@ -6,38 +6,94 @@ set -e
...
+declare -r ssh_host_key_dir="/.bottlerocket/host-containers/admin/etc/ssh"
+declare -r ssh_config_dir="/home/ec2-user/.ssh"
+declare -r user_data="/.bottlerocket/host-containers/admin/user-data"
+available_auth_methods=0
+
+parse_authorized_keys() {
+    # Extract the keys from user-data json
+    local raw_keys
+    if ! raw_keys=$(jq -e -r '.["ssh"]["authorized_keys"][]' "${user_data}" 2>/dev/null); then
+      echo "Failed to parse authorized_keys from ${user_data}" >&2
```
jahkeup: The `echo msg >&2` is repeated throughout, instead of relying on the convention of redirection a function would make the intent clear and safer overall. It would reduce the chances that extra output is sent to subshell substitutes that are expecting data in stdout.

```bash
log() {
  echo "$*" >&2
}
```

```suggestion
      log "Failed to parse authorized_keys from ${user_data}"
```


<details>
<summary>Or, alternatively (using my go to <samp>log</samp> definition)</summary>

```bash
# log with date prefix. 
# 
# Suggested use: `log $LEVEL: $msg`
log() {
  echo "[$(date --iso-8601=s --utc)] $*" >&2
}

# with use as:

log "INFO: installed ssh authorized keys"
log "WARN: unable to parse trusted user CA certificates"
log "ERROR: unable to setup ssh with provided user data"
```

</details>

start_admin_sshd.sh
```diff
@@ -6,38 +6,94 @@ set -e
...
-  if [[ ! "${public_key_data}" =~ ^"ssh" ]]; then
-    echo "Key ${public_key_data} with index ${public_key_index} looks invalid" >&2
-    continue
-  fi
-  echo "${public_key_data}" >> "${ssh_authorized_keys}"
-  if ! grep -q "${public_key_data}" "${ssh_authorized_keys}"; then
-    echo "Failed to write key with index ${public_key_index} to authorized_keys" >&2
-    continue
+declare -r ssh_host_key_dir="/.bottlerocket/host-containers/admin/etc/ssh"
+declare -r ssh_config_dir="/home/ec2-user/.ssh"
```
jahkeup: Let's put this above the `mkdir` and use it there to make sure we're being consistent throughout with directory creation and their early setup.

I'd also like to call this something else.. this isn't `/etc/ssh/ssh_config` (which would have been my guess once I lost this off my mental stack). May I suggest: `user_ssh_dir`. To me, this is far clearer in contexts like `"$user_ssh_dir/authorized_keys"` (vs. `"$ssh_config_dir/authorized_keys"`, which could have been a *system level* ssh config directory).

start_admin_sshd.sh
```diff
@@ -6,38 +6,94 @@ set -e
...
+
+  cat <<EOF >> "${sshd_config}"
+
+# Configured by user data
+TrustedUserCAKeys ${ssh_trusted_user_ca_keys}
+EOF
+
+  if [[ -f "${ssh_authorized_keys}" ]]; then
+    echo "AuthorizedKeysFile ${ssh_authorized_keys}" >> "${sshd_config}"
   fi
```
jahkeup: I don't think we should be configuring an absolute path to the authorized keys. I haven't tried this myself, but I'd suspect that an absolute path would authorize *all users* with the given file, meaning any user could login as any user. Even so, `sshd_config(5)` has this to say:


<pre>
AuthorizedKeysFile

  Specifies the file that contains the public keys used for user authentication.
  The format is described in the AUTHORIZED_KEYS FILE FORMAT section of sshd(8).
  Arguments to AuthorizedKeysFile accept the tokens described in the TOKENS
  section. After expansion, AuthorizedKeysFile is taken to be an absolute path
  or one relative to the user's home direc‐ tory. Multiple files may be listed,
  separated by whitespace. Alternately this option may be set to none to skip
  checking for user keys in files. 

  The default is ".ssh/authorized_keys .ssh/authorized_keys2".
</pre>

I'm of a mind that we do not try to set or modify the option. The location we're installing the keys will work without adding the explicit path.

start_admin_sshd.sh
```diff
@@ -6,38 +6,94 @@ set -e
...
-# If we didn't write any keys at all, there's not much point in continuing
-if [ ! -s "${ssh_authorized_keys}" ]; then
-  echo "Failed to write any valid public keys to authorized_keys" >&2
+  ((++available_auth_methods))
+fi
+
+# If there were no successful auth methods, then users cannot authenticate
+if [[ "${available_auth_methods}" -eq 0 ]]; then
+  echo "Failed to configure ssh authentication" >&2
   exit 1
```
jahkeup: The `${ssh_config_dir}` was created by the script, so we should make sure its `chown`'d before exiting. If users are doing anything to customize their image (and still using this script) then the directory may be inaccessible to them.

README.md
```diff
@@ -11,3 +11,33 @@ For more information about how the admin container fits into the Bottlerocket op
...
+To use custom public keys for `.ssh/authorized_keys` and/or custom CA keys for `/etc/ssh/trusted_user_ca_keys.pub` you will want to generate a JSON-structure like this:
+
+```
+{
+   "ssh":{
+      "authorized_keys":[
+         "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQClKsfkNkuSevGj3eYhCe53pcjqP3maAhDFcvBS7O6Vhz2ItxCih+PnDSUaw+WNQn/mZphTk/a/gU8jEzoOWbkM4yxyb/wB96xbiFveSFJuOp/d6RJhJOI0iBXrlsLnBItntckiJ7FbtxJMXLvvwJryDUilBMTjYtwB+QhYXUMOzce5Pjz5/i8SeJtjnV3iAoG/cQk+0FzZqaeJAAHco+CY/5WrUBkrHmFJr6HcXkvJdWPkYQS3xqC0+FmUZofz221CBt5IMucxXPkX4rWi+z7wB3RbBQoQzd8v7yeb7OzlPnWOyN0qFU0XA246RA8QFYiCNYwI3f05p6KLxEXAMPLE my-key-pair"
+      ],
+      "trusted_user_ca_keys":[
+         "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQClKsfkNkuSevGj3eYhCe53pcjqP3maAhDFcvBS7O6Vhz2ItxCih+PnDSUaw+WNQn/mZphTk/a/gU8jEzoOWbkM4yxyb/wB96xbiFveSFJuOp/d6RJhJOI0iBXrlsLnBItntckiJ7FbtxJMXLvvwJryDUilBMTjYtwB+QhYXUMOzce5Pjz5/i8SeJtjnV3iAoG/cQk+0FzZaeJAAHco+CY/5WrUBkrHmFJr6HcXkvJdWPkYQS3xqC0+FmUZofz221CBt5IMucxXPkX4rWi+z7wB3RbBQoQzd8v7yeb7OzlPnWOyN0qFU0XA246RA8QFYiCNYwI3f05p6KLxEXAMPLE my-key-pair"
```
jahkeup: I'd like to see a distinct and (maybe) valid example here instead of duplicated ssh public key (its the same as `authorized_keys`. At the very least, I think having a dummy value like `ssh-rsa example-ssh-ca-public-key-formatted-here authority@ssh-ca.example.com` is preferred.

jpculp:
- Switched `echos "blah" >&2` to `log()`
- Renamed `ssh_config_dir` to `user_ssh_dir`
- Replaced hardcoded `/home/ec2-user/.ssh` with `user_ssh_dir`
- Tweaked cat/EOF block
- Removed `echo "AuthorizedKeysFile ${ssh_authorized_keys}" >> "${sshd_config}"`
- Moved `chown ec2-user -R "${user_ssh_dir}"` up to run before the possible `exit` due to authentication configuration failure

README.md
```diff
@@ -11,3 +11,33 @@ For more information about how the admin container fits into the Bottlerocket op
...
+Users can add their own keys by populating the admin container's user-data with a base64-encoded JSON block.
+If user-data is populated then Bottlerocket will not fetch from IMDS at all, but if user-data is not set then Bottlerocket will continue to use the keys from IMDS.
+
+To use custom public keys for `.ssh/authorized_keys` and/or custom CA keys for `/etc/ssh/trusted_user_ca_keys.pub` you will want to generate a JSON-structure like this:
+
+```
+{
+   "ssh":{
+      "authorized_keys":[
+         "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQClKsfkNkuSevGj3eYhCe53pcjqP3maAhDFcvBS7O6Vhz2ItxCih+PnDSUaw+WNQn/mZphTk/a/gU8jEzoOWbkM4yxyb/wB96xbiFveSFJuOp/d6RJhJOI0iBXrlsLnBItntckiJ7FbtxJMXLvvwJryDUilBMTjYtwB+QhYXUMOzce5Pjz5/i8SeJtjnV3iAoG/cQk+0FzZqaeJAAHco+CY/5WrUBkrHmFJr6HcXkvJdWPkYQS3xqC0+FmUZofz221CBt5IMucxXPkX4rWi+z7wB3RbBQoQzd8v7yeb7OzlPnWOyN0qFU0XA246RA8QFYiCNYwI3f05p6KLxEXAMPLE my-key-pair"
```
bcressey: nit: might be good to give this the same `EXAMPLESSHPUBLICKEYFORMATTEDHERE` treatment since the line is very long, and we want to be clear that this isn't a valid key that will give us access to the instance of anyone who copy/pastes this.

start_admin_sshd.sh
```diff
@@ -4,64 +4,122 @@
...
+      valid_keys+=( "${key}" )
+    done
+
+    ( IFS=$'\n'; echo "${valid_keys[*]}" )
+}
+
+# Populate authorized_keys with all the authorized keys found in user-data
+if authorized_keys=$(parse_authorized_keys); then
+  ssh_authorized_keys="${user_ssh_dir}/authorized_keys"
+  install -D -m 0600 /dev/null "${ssh_authorized_keys}"
```
bcressey: This results in `~/.ssh` getting created with the default umask, but the recommended default is `0700`.

start_admin_sshd.sh
```diff
@@ -4,64 +4,122 @@
...
+    local -a valid_keys
+    for key in "${mapped_keys[@]}"; do
+      if ! echo "${key}" | ssh-keygen -lf - &>/dev/null; then
+        log "Failed to validate ${key}"
+        continue
+      fi
+      valid_keys+=( "${key}" )
+    done
+
+    ( IFS=$'\n'; echo "${valid_keys[*]}" )
```
bcressey, line 44: Can this common block be factored out into its own function?

start_admin_sshd.sh
```diff
@@ -4,64 +4,122 @@
...
+  ssh_authorized_keys="${user_ssh_dir}/authorized_keys"
+  install -D -m 0600 /dev/null "${ssh_authorized_keys}"
+  echo "${authorized_keys}" > "${ssh_authorized_keys}"
+  ((++available_auth_methods))
+fi
+
+parse_trusted_user_ca_keys() {
+    # Extract the keys from user-data json
+    local raw_keys
+    if ! raw_keys=$(jq -e -r '.["ssh"]["trusted_user_ca_keys"][]' "${user_data}" 2>/dev/null); then
```
bcressey: nit: I'd move this up by `parse_authorized_keys` so it doesn't interrupt the flow here.

start_admin_sshd.sh
```diff
@@ -4,64 +4,122 @@
...
+if trusted_user_ca_keys=$(parse_trusted_user_ca_keys); then
+  ssh_trusted_user_ca_keys="/etc/ssh/trusted_user_ca_keys.pub"
+  sshd_config="/etc/ssh/sshd_config"
+  install -D -m 0600 /dev/null "${ssh_trusted_user_ca_keys}"
+  echo "${trusted_user_ca_keys}" > "${ssh_trusted_user_ca_keys}"
+
+  cat >> "${sshd_config}" <<EOF
+
+# Configured by user data
+TrustedUserCAKeys ${ssh_trusted_user_ca_keys}
```
bcressey: Can this be set unconditionally and have `sshd` not complain too much if the file doesn't exist?

start_admin_sshd.sh
```diff
@@ -4,64 +4,122 @@
...
+EOF
+
+  ((++available_auth_methods))
 fi
 
-chown ec2-user -R "${ssh_config_dir}"
+chown ec2-user -R "${user_ssh_dir}"
```
bcressey: `ec2-user` should be in a variable so it's easier to override and matches how `user_ssh_dir` is defined

start_admin_sshd.sh
```diff
@@ -4,64 +4,122 @@
...
 mkdir -p "${ssh_host_key_dir}"
 for key in rsa ecdsa ed25519; do
-    # If both of the keys exist, don't overwrite them
-    if [ -s "${ssh_host_key_dir}/ssh_host_${key}_key" ] && [ -s "${ssh_host_key_dir}/ssh_host_${key}_key.pub"  ]; then
-        echo "${key} key already exists, will use existing key." >&2
-        continue
-    fi
+  # If both of the keys exist, don't overwrite them
+  if [[ -s "${ssh_host_key_dir}/ssh_host_${key}_key" ]] &&
+  [[ -s "${ssh_host_key_dir}/ssh_host_${key}_key.pub" ]]; then
```
bcressey: We don't need to use `[[` here; seems like an unnecessary change. I prefer to use `[` unless we're taking advantage of specific features that only `[[` can provide.

start_admin_sshd.sh
```diff
@@ -4,64 +4,122 @@
...
+  ((++available_auth_methods))
 fi
 
-chown ec2-user -R "${ssh_config_dir}"
+chown ec2-user -R "${user_ssh_dir}"
+
+# If there were no successful auth methods, then users cannot authenticate
+if [[ "${available_auth_methods}" -eq 0 ]]; then
+  log "Failed to configure ssh authentication"
+  exit 1
```
bcressey: We should just log a warning and start `sshd` anyhow, so the container doesn't continually exit and restart - there's no chance that a simple restart will clear up a transient failure.

README.md
jpculp: You're right, it does look a little silly now. I'll change it to be more consistent.

Fun fact, the example key is from the IMDS documentation. (for anyone who is curious)

start_admin_sshd.sh
jpculp: Based on the other feedback, I'll create a single function to take care of both 😃

jahkeup (approved):

jpculp:
- Consolidated `parse_authorized_keys/parse_trusted_user_ca_keys` into a single `get_user_data_keys` function.
- Reduced example authorized_key to something shorter.
- Added `TrustedUserCAKeys` directly to `sshd_config`.
- Replaced hardcoded `ec2-user` with variable `local_user`
- Replaced `install` with the original `touch/chmod` to keep things simple.
- Removed extraneous brackets.

zmrow (approved):
:video_game:

etungsten (approved):

webern (approved):

bcressey (approved):

PR #20, opened 2021-03-30
Improve error handling and logging during ssh setup

jpculp:
**Issue number:**

N/A

**Description of changes:**

This adds exits on failure to two configuration checkpoints in the `start_admin_sshd.sh` script in order to prevent an improperly configured ssh daemon from starting. Prior to this, an error would be printed, but the script would continue to run.

This also replaces a forward of `stderr` to `/dev/null` when getting `user-data` keys and replaces the in-function error message with a more contextual message when there are no available auth methods. The available auth methods check error will output the `user-data` JSON to help with troubleshooting.

These changes will, among other things, prevent this common error message printed to the EC2 console when a user relies on `authorized_keys` as opposed to `trusted_user_ca_keys`:
```
host-ctr[905]: Failed to parse trusted_user_ca_keys from /.bottlerocket/host-containers/admin/user-data
```

**Testing done:**

- Launched publicly available `aws-ecs-1` ami.
- Connected to control container via ssm session.
- Changed the source for the admin container and enabled it.

After that, I went through several test cases of disabling the admin container, posting a new base64-encoded block to `host-containers.admin.user-data`,  re-enabling the admin container, and checking the connection.
- :white_check_mark: Empty JSON Structure (expected fail)
- :white_check_mark: Custom authorized key
- :white_check_mark: Several keys, some good, some bad (only the strings that could be parsed by ssh-keygen were present in `authorized_keys`)
- :white_check_mark: Authorized key AND trusted user CA key (`.ssh/authorized_keys`, `/etc/ssh/trusted_user_ca_keys.pub`, and `/etc/ssh/sshd_config` looked as expected)

**Terms of contribution:**

### DIFF
--- a/start_admin_sshd.sh
+++ b/start_admin_sshd.sh
@@ -21,8 +21,10 @@ get_user_data_keys() {
     # Extract the keys from user-data json
     local raw_keys
     local key_type="${1:?}"
-    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]' "${user_data}" 2>/dev/null); then
-      log "Failed to parse ${key_type} from ${user_data}"
+    # ? signifies an optional object identifier-index and doesn't return a 'failed to iterate' error
+    # in the event that a key_type is missing. If jq returns nothing, then that is caught by the -z.
+    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${user_data}") \
+    || [[ -z "${raw_keys}" ]]; then
       return 1
     fi
 
@@ -66,7 +68,8 @@ chown "${local_user}" -R "${user_ssh_dir}"
 
 # If there were no successful auth methods, then users cannot authenticate
 if [[ "${available_auth_methods}" -eq 0 ]]; then
-  log "Failed to configure ssh authentication"
+  user_data_condensed=$(tr -d '[:space:]' < "${user_data}")
+  log "Failed to configure ssh authentication with user-data: ${user_data_condensed}"
   exit 1
 fi
 
--- a/start_admin_sshd.sh
+++ b/start_admin_sshd.sh
@@ -67,6 +67,7 @@ chown "${local_user}" -R "${user_ssh_dir}"
 # If there were no successful auth methods, then users cannot authenticate
 if [[ "${available_auth_methods}" -eq 0 ]]; then
   log "Failed to configure ssh authentication"
+  exit 1
 fi
 
 # Generate the server keys
@@ -87,6 +88,7 @@ for key in rsa ecdsa ed25519; do
     chmod 644 "${ssh_host_key_dir}/ssh_host_${key}_key.pub"
   else
     log "Failure to generate host ${key} ssh keys"
+    exit 1
   fi
 done
 


start_admin_sshd.sh
```diff
@@ -21,8 +21,10 @@ get_user_data_keys() {
     # Extract the keys from user-data json
     local raw_keys
     local key_type="${1:?}"
-    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]' "${user_data}" 2>/dev/null); then
-      log "Failed to parse ${key_type} from ${user_data}"
+    # ? signifies an optional object identifier-index and doesn't return a 'failed to iterate' error
+    # in the event that a key_type is missing. If jq returns nothing, then that is caught by the -z.
+    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${user_data}") \
+    || [[ -z "${raw_keys}" ]]; then
```
etungsten, line 27: I'm pretty sure you can use the `-e` flag for that.

> -e / --exit-status:

           Sets  the  exit  status of jq to 0 if the last output values
           was neither false nor null, 1 if the last output  value  was
           either  false or null, or 4 if no valid result was ever pro‐
           duced. Normally jq exits with 2 if there was any usage prob‐
           lem  or  system  error,  3 if there was a jq program compile
           error, or 0 if the jq program ran.

```suggestion
    if ! raw_keys=$(jq -e --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${user_data}"); then
```

jpculp: When I was testing I found that jq actually will return a 0 on an empty list, since the list exists but just happens to not be populated.

jpculp: After some further testing, I verified the -z is not necessary.

jpculp:
- Removed the `|| [[ -z "${raw_keys}" ]]` logic since `jq` still returns `4` even with the `?`, and the comment that referenced it.

webern (approved):

etungsten (approved):

tjkirch:
>     * Removed the `|| [[ -z "${raw_keys}" ]]` logic since `jq` still returns `4` even with the `?`, and the comment that referenced it.

I think it's a good check to have in general.  If someone mistakenly puts in an empty entry, for example with a typo of a variable, an unexpected input to their data, etc. then it should still fail.  As it is, this passes: `echo '{"ssh": {"authorized_keys": [""]}}' | jq --arg key_type "authorized_keys" -e -r '.["ssh"][$key_type][]?'`

jpculp:
- Reverted the last force-push to re-include the empty key check based on @tjkirch's feedback

tjkirch (approved):

etungsten (approved):

bcressey (approved):

zmrow (approved):
:wolf:

start_admin_sshd.sh
bcressey, line 27: nit: I'd prefer a less complex conditional check; assigning a variable and then checking it in a subsequent test is a bit too clever for my taste.

```suggestion
    raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${user_data}")
    if [ "${?}" -gt 0 ] || [ -z "${raw_keys}" ] ; then
```

tjkirch: That doesn't work with `set -e`; the shell would exit after the assignment before it can even get to the `$?` check.

bcressey (approved):

PR #22, opened 2021-03-31
Clean up start_admin_sshd.sh

jpculp:
**Issue number:**

N/A

**Description of changes:**

This is just housekeeping to be more consistent with control container and to increase readability.
There are no functional changes.

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/start_admin_sshd.sh
+++ b/start_admin_sshd.sh
@@ -1,31 +1,34 @@
-#!/bin/bash
-# This file is part of Bottlerocket.
-# Copyright Amazon.com, Inc., its affiliates, or other contributors. All Rights Reserved.
-# SPDX-License-Identifier: Apache-2.0 OR MIT
+#!/usr/bin/env bash
+
 set -e
+
+declare -r PERSISTENT_STORAGE_BASE_DIR="/.bottlerocket/host-containers/current"
+declare -r SSH_HOST_KEY_DIR="${PERSISTENT_STORAGE_BASE_DIR}/etc/ssh"
+declare -r USER_DATA="${PERSISTENT_STORAGE_BASE_DIR}/user-data"
+
+declare -r LOCAL_USER="ec2-user"
+declare -r USER_SSH_DIR="/home/${LOCAL_USER}/.ssh"
+
+# This is a counter used to verify at least
+# one of the methods below is available.
+declare -i available_auth_methods=0
 
 log() {
   echo "$*" >&2
 }
 
-declare -r local_user="ec2-user"
-declare -r persistent_storage_base_dir="/.bottlerocket/host-containers/current"
-declare -r ssh_host_key_dir="${persistent_storage_base_dir}/etc/ssh"
-declare -r user_data="${persistent_storage_base_dir}/user-data"
-declare -r user_ssh_dir="/home/${local_user}/.ssh"
-available_auth_methods=0
-
-mkdir -p "${user_ssh_dir}"
-chmod 700 "${user_ssh_dir}"
-
 get_user_data_keys() {
     # Extract the keys from user-data json
     local raw_keys
     local key_type="${1:?}"
-    # ? signifies an optional object identifier-index and doesn't return a 'failed to iterate' error
-    # in the event that a key_type is missing. If jq returns nothing, then that is caught by the -z.
-    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${user_data}") \
-    || [[ -z "${raw_keys}" ]]; then
+    # ? signifies an optional object identifier-index and doesn't return a
+    # 'failed to iterate' error in the event that a key_type is missing.
+    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${USER_DATA}"); then
+      return 1
+    fi
+
+    # Verify jq returned key(s)
+    if [[ -z "${raw_keys}" ]]; then
       return 1
     fi
 
@@ -58,16 +61,19 @@ EOF
   chmod 644 "${proxy_profile}"
 }
 
+mkdir -p "${USER_SSH_DIR}"
+chmod 700 "${USER_SSH_DIR}"
+
 # Populate authorized_keys with all the authorized keys found in user-data
 if authorized_keys=$(get_user_data_keys "authorized_keys"); then
-  ssh_authorized_keys="${user_ssh_dir}/authorized_keys"
+  ssh_authorized_keys="${USER_SSH_DIR}/authorized_keys"
   touch "${ssh_authorized_keys}"
   chmod 600 "${ssh_authorized_keys}"
   echo "${authorized_keys}" > "${ssh_authorized_keys}"
   ((++available_auth_methods))
 fi
 
-# Populate trusted_user_ca_keys with all the authorized keys found in user-data
+# Populate trusted_user_ca_keys with all the trusted ca keys found in user-data
 if trusted_user_ca_keys=$(get_user_data_keys "trusted_user_ca_keys"); then
   ssh_trusted_user_ca_keys="/etc/ssh/trusted_user_ca_keys.pub"
   touch "${ssh_trusted_user_ca_keys}"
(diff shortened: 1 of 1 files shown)


start_admin_sshd.sh
```diff
@@ -47,9 +35,23 @@ get_user_data_keys() {
...
 
+declare -r PERSISTENT_STORAGE_BASE_DIR="/.bottlerocket/host-containers/current"
+declare -r SSH_HOST_KEY_DIR="${PERSISTENT_STORAGE_BASE_DIR}/etc/ssh"
+declare -r USER_DATA="${PERSISTENT_STORAGE_BASE_DIR}/user-data"
+
+declare -r LOCAL_USER="ec2-user"
+declare -r USER_SSH_DIR="/home/${LOCAL_USER}/.ssh"
+
+# This is a counter used to verify at least
+# one of the methods of below is available
```
webern: Maybe you meant this?

```suggestion
# one of the methods below is available
```

webern (approved):

jpculp:
- Fixed typo

start_admin_sshd.sh
```diff
@@ -65,31 +67,31 @@ if trusted_user_ca_keys=$(get_user_data_keys "trusted_user_ca_keys"); then
   ((++available_auth_methods))
 fi
 
-chown "${local_user}" -R "${user_ssh_dir}"
+chown "${LOCAL_USER}" -R "${USER_SSH_DIR}"
 
 # If there were no successful auth methods, then users cannot authenticate
 if [[ "${available_auth_methods}" -eq 0 ]]; then
-  user_data_condensed=$(tr -d '[:space:]' < "${user_data}")
+  user_data_condensed=$(tr -d '[:space:]' < "${USER_DATA}")
```
jahkeup, line 89: This command would delete *intentional* whitespace as well as the (potential) extraneous space from read JSON. If the goal is to compact the JSON itself (and not user values), then let's use `jq` to do so - we already know its there and it has a handy flag for this:

```bash
# print out condensed json blob to logs.
jq --compact-output . "$USER_DATA"

# or handle fallback, when not json (I didn't read back to see if we 
# return early away from this path if its not known to be json):
jq --exit-status --compact-output . "$USER_DATA" 2>/dev/null || cat "$USER_DATA"
```

start_admin_sshd.sh
```diff
@@ -65,31 +67,31 @@ if trusted_user_ca_keys=$(get_user_data_keys "trusted_user_ca_keys"); then
   ((++available_auth_methods))
 fi
 
-chown "${local_user}" -R "${user_ssh_dir}"
+chown "${LOCAL_USER}" -R "${USER_SSH_DIR}"
```
jahkeup: nit: flag before the positional arguments is preferred. Added `:` after the local user to discourage misusing `LOCAL_USER=ec2-user:myneatgroup`

```suggestion
chown -R "${LOCAL_USER}:" "${USER_SSH_DIR}"
```

start_admin_sshd.sh
```diff
@@ -65,31 +67,31 @@ if trusted_user_ca_keys=$(get_user_data_keys "trusted_user_ca_keys"); then
...
-  user_data_condensed=$(tr -d '[:space:]' < "${user_data}")
+  user_data_condensed=$(tr -d '[:space:]' < "${USER_DATA}")
   log "Failed to configure ssh authentication with user-data: ${user_data_condensed}"
   exit 1
 fi
 
 # Generate the server keys
-mkdir -p "${ssh_host_key_dir}"
+mkdir -p "${SSH_HOST_KEY_DIR}"
 for key in rsa ecdsa ed25519; do
```
jahkeup: Please rename this var while you're here. These aren't "keys" but rather a list of "key algorithms" to generate:

```suggestion
for keyalg in rsa ecdsa ed25519; do
```

start_admin_sshd.sh
```diff
@@ -47,9 +35,23 @@ get_user_data_keys() {
...
+declare -r PERSISTENT_STORAGE_BASE_DIR="/.bottlerocket/host-containers/current"
+declare -r SSH_HOST_KEY_DIR="${PERSISTENT_STORAGE_BASE_DIR}/etc/ssh"
+declare -r USER_DATA="${PERSISTENT_STORAGE_BASE_DIR}/user-data"
+
+declare -r LOCAL_USER="ec2-user"
+declare -r USER_SSH_DIR="/home/${LOCAL_USER}/.ssh"
+
+# This is a counter used to verify at least
+# one of the methods below is available
+available_auth_methods=0
```
jahkeup: Please declare this as an integer to avoid introducing bugs in later revisions: 

```suggestion
declare -i available_auth_methods=0
```

Example of why this is important:

```bash
#!/usr/bin/env bash
unset fooint foostr
declare -i fooint=0
foostr=0

fooint+=1
foostr+=1

echo "fooint: $fooint, foostr: $foostr"

#stdout: fooint: 1, foostr: 01
```

etungsten (approved):

start_admin_sshd.sh
```diff
@@ -1,30 +1,18 @@
...
-
 get_user_data_keys() {
     # Extract the keys from user-data json
     local raw_keys
     local key_type="${1:?}"
     # ? signifies an optional object identifier-index and doesn't return a 'failed to iterate' error
     # in the event that a key_type is missing. If jq returns nothing, then that is caught by the -z.
-    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${user_data}") \
+    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${USER_DATA}") \
     || [[ -z "${raw_keys}" ]]; then
```
jahkeup: Rather than combining this in the same if, can you split this out into its own branch?

I've added some logging messages to help demonstrate the difference, not necessarily to suggest adding the messages:

```bash
if ! raw_keys=$(jq ...); then
  # can't extract keys from provided data
  log "unable to parse configuration from user-data"
  return 1
fi

if [[ -z "$raw_keys" ]]; then
  # no keys defined
  log "user-data does not provide $raw_keys"
  return 1
fi
```

start_admin_sshd.sh
```diff
@@ -47,9 +35,23 @@ get_user_data_keys() {
     ( IFS=$'\n'; echo "${valid_keys[*]}" )
 }
 
+declare -r PERSISTENT_STORAGE_BASE_DIR="/.bottlerocket/host-containers/current"
+declare -r SSH_HOST_KEY_DIR="${PERSISTENT_STORAGE_BASE_DIR}/etc/ssh"
+declare -r USER_DATA="${PERSISTENT_STORAGE_BASE_DIR}/user-data"
```
jahkeup: It's a bit odd to me to see this declared *below* a user of the variable - was this organized this way intentionally? What your thought on this?

jahkeup (changes requested):

start_admin_sshd.sh
jpculp: I thought it would easier to read if the functions were at the top and the main logic was below (including constants as part of the main logic)

start_admin_sshd.sh
jpculp: That's fair. I'll definitely make this change, but I'd like to do so in a different PR to keep this limited to non-functional changes.

start_admin_sshd.sh
jahkeup: I'd prefer that we had these constants up top. 

There aren't user input changes to them (no flag parsing) or otherwise so they may as well be declared early on. It also loads up the mental context with the constants prior to reading their uses.

There's not too much code and assumption between the functions and constants at this point so I'll leave it up to you.

jpculp: Sure, I'll go ahead and make this change here, and in the control container script as well since my [PR](https://github.com/bottlerocket-os/bottlerocket-control-container/pull/12) is still open.

jpculp:
- Moved constants to the top
- Set `declare -i` for `available_auth_methods`
- Split up the `raw_keys` jq logic
- Fixed the chown
- Renamed `key` as `key_alg`

webern (approved):

jpculp:
- Switched `#!/bin/bash` to `#!/usr/bin/env bash`
- Fixed typo in an old comment

start_admin_sshd.sh
```diff
@@ -1,31 +1,34 @@
...
     # Extract the keys from user-data json
     local raw_keys
     local key_type="${1:?}"
-    # ? signifies an optional object identifier-index and doesn't return a 'failed to iterate' error
-    # in the event that a key_type is missing. If jq returns nothing, then that is caught by the -z.
-    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${user_data}") \
-    || [[ -z "${raw_keys}" ]]; then
+    # ? signifies an optional object identifier-index and doesn't return a
+    # 'failed to iterate' error in the event that a key_type is missing.
+    if ! raw_keys=$(jq --arg key_type "${key_type}" -e -r '.["ssh"][$key_type][]?' "${USER_DATA}"); then
```
jahkeup, line 26: I'm getting to the point where I think we ought to add tests in the very very near future (not this PR; maybe with [BATS](https://github.com/bats-core/bats-core)?) for the scripts in the host containers. The functionality is steadily increasing that having some unit tests for processing functions would be great to ensure expected cases are handled appropriately.

start_admin_sshd.sh
```diff
@@ -65,33 +71,33 @@ if trusted_user_ca_keys=$(get_user_data_keys "trusted_user_ca_keys"); then
   ((++available_auth_methods))
```
jahkeup, line 82: Oh, now that the counter is an int, these can become: `available_auth_methods+=1`. This isn't necessary, but it is a related cleanup item!

jahkeup (approved):
🎉

start_admin_sshd.sh
jpculp: I think we tried this back when we first added it, but it didn't cooperate with `set -e`. At least I remember that being the case with `((available_auth_methods++))`. If you're not strongly opposed I'd rather keep as is.

start_admin_sshd.sh
jpculp: I'm a fan of automated testing 🤖

start_admin_sshd.sh
jahkeup: We tried `((inside++))` which does exit non zero on 0->1 transitions - this behaves coherently:

```bash
trap 'echo exit=$? foo=$foo' DEBUG

declare -i foo=0
false # the only non-zero exit below
foo+=1
foo+=1
echo foo=$foo # should be 2!

#stdout: exit=0 foo=
#stdout: exit=0 foo=0
#stdout: exit=1 foo=0
#stdout: exit=0 foo=1
#stdout: exit=0 foo=2
#stdout: foo=2
```

jahkeup: Also: fine to leave as is 👍🏽

jpculp: Ah nifty! I'm still a fan of the `++`, but this gave me a stronger appreciation for the `declare -i`. I'll be sure to keep that up in the future.

etungsten (approved):

jpculp:
- Rebased to fix merge-conflict

start_admin_sshd.sh
jpculp: Addressed in #25

PR #24, opened 2021-03-31
start sshd script: export BR proxy env vars for all login shells

### LINKED ISSUE #23: write proxy env vars into the default user environment files under `/etc`
> Note that the default admin container's entry point is start_admin_sshd.sh and at the end of the script it execs the sshd daemon and proxy environment variables won't be present in the sshd spawned shells

We'd need to fix in the admin container by writing these values into the default user environment files under `/etc`.

_Originally posted by @bcressey in https://github.com/bottlerocket-os/bottlerocket/issues/1432#issuecomment-810659284_

etungsten:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket-admin-container/issues/23

**Description of changes:**
```
Author: Erikson Tung <etung@amazon.com>
Date:   Wed Mar 31 12:36:44 2021 -0700

    start sshd script: export BR proxy env vars for all login shells

```

**Testing done:**
Launch instance with AMI that has changes from https://github.com/bottlerocket-os/bottlerocket/pull/1432, with the proxy settings in my userdata:
```
[settings.network]
https-proxy="somewhere:9898"
no-proxy = [".local"]
```

SSH'd to my admin container and saw proxy env vars exported successfully:
```
[ec2-user@ip-192-168-8-185 ~]$ echo $HTTPS_PROXY 
somewhere:9898
[ec2-user@ip-192-168-8-185 ~]$ echo $NO_PROXY 
.local,localhost,127.0.0.1,191bc2a1c5d1e9e863c973a7bdfb88d0.yl4.us-west-2.eks.amazonaws.com,.cluster.local
[ec2-user@ip-192-168-8-185 ~]$ echo $no_proxy 
.local,localhost,127.0.0.1,191bc2a1c5d1e9e863c973a7bdfb88d0.yl4.us-west-2.eks.amazonaws.com,.cluster.local
[ec2-user@ip-192-168-8-185 ~]$ echo $https_proxy 
somewhere:9898
```

Then I changed my NO_PROXY setting, restarted the admin host-container:
```
bash-5.0# apiclient set --json '{"network": { "no-proxy": ["hello-there"]}}'
bash-5.0# systemctl restart host-containers@admin
```
Logged back into my admin container and saw the changes to proxy env var reflected in the login shell:
```
[ec2-user@ip-192-168-8-185 ~]$ echo $NO_PROXY 
hello-there,localhost,127.0.0.1,191bc2a1c5d1e9e863c973a7bdfb88d0.yl4.us-west-2.eks.amazonaws.com,.cluster.local
```

**Terms of contribution:**

### DIFF
--- a/start_admin_sshd.sh
+++ b/start_admin_sshd.sh
@@ -47,6 +47,17 @@ get_user_data_keys() {
     ( IFS=$'\n'; echo "${valid_keys[*]}" )
 }
 
+# Export proxy environment variables for all users' login shells
+# Match the values of the proxy environment variables given to the admin container
+install_proxy_profile() {
+  local -r proxy_profile="/etc/profile.d/bottlerocket-proxy-settings.sh"
+  cat > "${proxy_profile}" <<EOF
+# Export Bottlerocket proxy environment variables
+$(declare -p HTTP_PROXY HTTPS_PROXY no_proxy NO_PROXY 2>/dev/null)
+EOF
+  chmod 644 "${proxy_profile}"
+}
+
 # Populate authorized_keys with all the authorized keys found in user-data
 if authorized_keys=$(get_user_data_keys "authorized_keys"); then
   ssh_authorized_keys="${user_ssh_dir}/authorized_keys"
@@ -95,6 +106,8 @@ for key in rsa ecdsa ed25519; do
     exit 1
   fi
 done
+
+install_proxy_profile
 
 # Start a single sshd process in the foreground
 exec /usr/sbin/sshd -e -D


start_admin_sshd.sh
```diff
@@ -96,5 +96,22 @@ for key in rsa ecdsa ed25519; do
   fi
 done
 
+# Export proxy environment variables for all users' login shells
+# Match the values of the proxy environment variables given to the admin container
+PROXY_ENV_VAR_SETUP_SCRIPT=/etc/profile.d/br-proxy-env.sh
```
jpculp: nit. the other variables in the script have their values wrapped in quotes

start_admin_sshd.sh
```diff
@@ -96,5 +96,22 @@ for key in rsa ecdsa ed25519; do
...
+if [ ! -z ${HTTPS_PROXY:+x} ]; then
```
jpculp: I would recommend getting in the habit of using `[[..]]` over `[..]`, but its one of those things with passionate people on both sides. For consistency it would be nice if we could standardize on one or the other at least for the control and admin containers.

jpculp (commented):
Given that this is kind of it's own thing while the rest of the script is `sshd`-centric, should this just be it's own function defined near the top of the script?

etungsten:
Push above addresses @jpculp 's comments. Tested again and things still work as expected.

etungsten:
Push above removes an extraneous empty line in the heredoc

start_admin_sshd.sh
```diff
@@ -47,6 +47,26 @@ get_user_data_keys() {
     ( IFS=$'\n'; echo "${valid_keys[*]}" )
 }
 
+# Export proxy environment variables for all users' login shells
+# Match the values of the proxy environment variables given to the admin container
+set_up_proxy_env_vars() {
+  PROXY_ENV_VAR_SETUP_SCRIPT="/etc/profile.d/br-proxy-env.sh"
```
jahkeup: Please declare this as a constant:

```suggestion
  local -r PROXY_ENV_VAR_SETUP_SCRIPT="/etc/profile.d/br-proxy-env.sh"
```

start_admin_sshd.sh
```diff
@@ -47,6 +47,26 @@ get_user_data_keys() {
...
+  local export_https_proxy
+  local export_no_proxy
+  if [[ ! -z ${HTTPS_PROXY:+x} ]]; then
+    export_https_proxy="export HTTPS_PROXY=${HTTPS_PROXY} https_proxy=${HTTPS_PROXY}"
+  fi
+  if [[ ! -z ${NO_PROXY:+x} ]]; then
+    export_no_proxy="export NO_PROXY=${NO_PROXY} no_proxy=${NO_PROXY}"
+  fi
```
jahkeup: The if-present substitution seems redundant - directly checking the non-empty state is clearer here:

```suggestion
  if [[ -n "$HTTPS_PROXY" ]]; then
    export_https_proxy="export HTTPS_PROXY=${HTTPS_PROXY} https_proxy=${HTTPS_PROXY}"
  fi
  if [[ -n "$NO_PROXY" ]]; then
    export_no_proxy="export NO_PROXY=${NO_PROXY} no_proxy=${NO_PROXY}"
  fi
```

start_admin_sshd.sh
```diff
@@ -47,6 +47,26 @@ get_user_data_keys() {
...
+  cat <<EOF > ${PROXY_ENV_VAR_SETUP_SCRIPT}
+# Export Bottlerocket proxy environment variables
+${export_https_proxy}
+${export_no_proxy}
+EOF
```
jahkeup: Instead of accumulating the variables with string built commands, this can instead be driven with a "safer" use of `declare -p`:

```bash
cat <<EOF
# Bottlerocket proxy settings
$(declare -p HTTP_PROXY HTTPS_PROXY no_proxy NO_PROXY 2>/dev/null)
EOF
```
<details>
<summary>Example of this used</summary>

```bash
( set -x; export HTTP_PROXY="http://myproxy.example.com/cool/url that's awesome"; export HTTPS_PROXY="https://myproxy.example.com/really/awesome url?ruok=irok"; unset no_proxy; echo "# declare -p ..."; declare -p HTTP_PROXY HTTPS_PROXY http_proxy https_proxy no_proxy; echo "# stderr redirected:"; declare -p HTTP_PROXY HTTPS_PROXY http_proxy https_proxy no_proxy 2>/dev/null; )
```

```bash
# declare -p ...
declare -x HTTP_PROXY="http://myproxy.example.com/cool/url that's awesome"
declare -x HTTPS_PROXY="https://myproxy.example.com/really/awesome url?ruok=irok"
bash: declare: http_proxy: not found
bash: declare: https_proxy: not found
bash: declare: no_proxy: not found
# stderr redirected:
declare -x HTTP_PROXY="http://myproxy.example.com/cool/url that's awesome"
declare -x HTTPS_PROXY="https://myproxy.example.com/really/awesome url?ruok=irok"
```
</details>

start_admin_sshd.sh
jpculp: Sorry I didn't notice this before, but could you set this to `local -r proxy_env_var_setup_script` or `declare -r PROXY_ENV_VAR_SETUP_SCRIPT` outside of the function and with the other constants?

start_admin_sshd.sh
```diff
@@ -47,6 +47,26 @@ get_user_data_keys() {
...
+  chmod 644 ${PROXY_ENV_VAR_SETUP_SCRIPT}
```
jpculp: super nit. it might be nice to wrap this is quotes, but its not required.

start_admin_sshd.sh
```diff
@@ -47,6 +47,26 @@ get_user_data_keys() {
     ( IFS=$'\n'; echo "${valid_keys[*]}" )
 }
 
+# Export proxy environment variables for all users' login shells
+# Match the values of the proxy environment variables given to the admin container
+set_up_proxy_env_vars() {
```
jahkeup: nit: I suggest naming this to align with its actions and the role it plays - installing a profile script that included the proxy details for use by shells.

```suggestion
install_proxy_profile() {
```

jahkeup (changes requested):

start_admin_sshd.sh
jpculp: `-n` is also recommended by [SC2236](https://github.com/koalaman/shellcheck/wiki/SC2236)

etungsten: I wanted to differentiate between the environment variable being unset, and being set but zero-length. I don't want to export any of the proxy variables if they were unset.

etungsten: See https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
> ${parameter:+word}
 >  If parameter is null or unset, nothing is substituted, otherwise the expansion of word is substituted.


`-n` can't check unset values, it'll always return true. Since it'll expand to just `[ -n ];`

The desired effect is for the condition to evaluate to true if the environment is _set_ (not necessary length > 0)

etungsten: This is no longer necessary if I use `declare -x` instead of exports as described in https://github.com/bottlerocket-os/bottlerocket-admin-container/pull/24#discussion_r605264971

jahkeup: > -n can't check unset values, it'll always return true. Since it'll expand to just [ -n ];

This doesn't apply to `[[`. One of the reasons why I'm a proponent of `[[` 😄 .

etungsten:
Push above addresses comments by @jahkeup and @jpculp . Tested things and they still work as expected.

jpculp (approved):

samuelkarp (approved):

webern (approved):

start_admin_sshd.sh
```diff
@@ -47,6 +47,17 @@ get_user_data_keys() {
     ( IFS=$'\n'; echo "${valid_keys[*]}" )
 }
 
+# Export proxy environment variables for all users' login shells
+# Match the values of the proxy environment variables given to the admin container
+install_proxy_profile() {
+  local -r PROXY_ENV_VAR_SETUP_SCRIPT="/etc/profile.d/br-proxy-env.sh"
+  cat <<EOF > ${PROXY_ENV_VAR_SETUP_SCRIPT}
```
jahkeup: I think we lose clarity for the sake of brevity here. What do you think about calling this: 

```suggestion
  local -r proxy_profile="/etc/profile.d/bottlerocket-proxy-settings.sh"
  cat > "$proxy_profile" <<EOF
```

jahkeup (approved):
Minor suggestion, otherwise lgtm!

etungsten:
Push above address the comment by @jahkeup 

Tested again and things still work as expected.

jahkeup (approved):

PR #25, opened 2021-04-01
Fix condensed user data

jpculp:
**Issue number:**

N/A

**Description of changes:**

This replaces the `tr -d '[:space:]'` which removed all whitespace, including intentional whitespace.
`jq -c` compacts output to delete extraneous spaces without removing intentional ones.
If the user data is not JSON, then we will simply `cat` out the file.

**Testing done:**

Without specified user-data:

With purposefully malformed user data
- Launched instance with `user-data` set to base64-encoded "Hello world"
- Verified in the EC2 console that the expected errors appeared and printed "Hello world"
- Launched instance with mostly correct `user-data` but `authorized_keyz` instead of `authorized_keys`
- Verified in the EC2 console that the error showed the JSON without extraneous spaces, but preserving the one between the key and the host that generated it. 

**Terms of contribution:**

### DIFF
--- a/start_admin_sshd.sh
+++ b/start_admin_sshd.sh
@@ -86,7 +86,7 @@ chown -R "${LOCAL_USER}:" "${USER_SSH_DIR}"
 
 # If there were no successful auth methods, then users cannot authenticate
 if [[ "${available_auth_methods}" -eq 0 ]]; then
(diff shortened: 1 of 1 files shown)


etungsten (approved):

jahkeup (approved):
🗜️ yeslooksgreattome!

zmrow (approved):
:socks:

PR #26, opened 2021-04-02
Add kebab-case settings as the default

jpculp:
**Issue number:**

N/A

**Description of changes:**

This adds support for settings in kebab-case which is more consistent with Bottlerocket's API.
For legacy compatibility, we have added an OR conditional for the snake_case equivalents from v0.6.0.

**Testing done:**

Then I reran the same test as above with an ami containing a [kebab-case patched shibaken](https://github.com/bottlerocket-os/bottlerocket/pull/1439).

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,10 @@ To use custom public keys for `.ssh/authorized_keys` and/or custom CA keys for `
 ```
 {
    "ssh":{
-      "authorized_keys":[
+      "authorized-keys":[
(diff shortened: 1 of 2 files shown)


bcressey (approved):

etungsten (approved):

tjkirch (approved):

zmrow (approved):
:headphones:

PR #35, opened 2021-05-20
Pull bash and musl sources from lookaside cache

jpculp:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket-admin-container/issues/1

**Description of changes:**

This commit pulls and verifies sources from the Bottlerocket SDK's "lookaside-cache", as opposed to their upstream repositories.

**Testing done:**

- Built the container.
- Pushed to ECR.
- Ran `$0 --version` to verify correct build version **5.0.18**.

**Terms of contribution:**


sdk-fetch
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
jpculp, line 1: We've started to use `#!/usr/bin/env bash`, but I left this untouched from [bottlerocket-sdk/sdk-fetch](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/develop/sdk-fetch). Should we just switch both?

jpculp: Switching to `#!/usr/bin/env bash` in [bottlerocket-sdk/sdk-fetch](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/develop/sdk-fetch) seemed to cause issues. I'll keep this as is, at least for now.

etungsten (approved):

webern (approved):

PR #37, opened 2021-06-09
Add helper to run processes in other host containers

tjkirch:
**Issue number:**

This is a partial solution to https://github.com/bottlerocket-os/bottlerocket/issues/367 that allows you to access other host containers from the admin container.  (A full solution would allow access from any host container with API access, which has some design challenges, but this simple helper eases some currently ugly use cases around debugging host containers.)

**Description of changes:**

Adds a `collie` helper, similar to `sheltie`, that `coll`s programs in other host containers using nsenter and `ctr exec`.

**Testing done:**

The default case gives you bash, with a TTY:
```
$ sudo collie admin
bash-4.2#
```

You can run other processes; I imagine a common use will be looking at files, and you can redirect them to a file to copy them out.
```
$ sudo collie control head -n1 /etc/motd
Welcome to Bottlerocket's control container!
```

(A number of other commands work as expected, too.)

Error from a nonexistent container:
```
$ sudo collie alsdkjf
ctr: container "alsdkjf" in namespace "default": not found
```

Usage message without arguments:
```
$ sudo collie
Must specify host container name.

Usage: collie [ --tty | -t ] CONTAINER [ COMMAND [ ARGS ] ]

Runs the given command inside the given host container.

COMMAND defaults to /bin/bash.
```

Usage message without sudo:
```
$ collie control
collie must be run as root, try 'sudo collie'
```

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,11 +59,10 @@ ADD --chown=root:root motd /etc/
 ADD --chown=root:root ec2-user.sudoers /etc/sudoers.d/ec2-user
 ADD start_admin_sshd.sh /usr/sbin/
 ADD ./sshd_config /etc/ssh/
-ADD ./sheltie /usr/bin/
+ADD ./sheltie ./collie /usr/bin/
 
 RUN chmod 440 /etc/sudoers.d/ec2-user
-RUN chmod +x /usr/sbin/start_admin_sshd.sh
-RUN chmod +x /usr/bin/sheltie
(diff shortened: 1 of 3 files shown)


jpculp (approved):
:dog2:

zmrow (approved):
Nice work! :zap:

tjkirch:
I'm hearing too much consternation about not being able to call this from other host containers. Closing.

Issue #38, opened 2021-06-28
add support for EC2 Instance Connect


bcressey:
**Image I'm using:**
v0.7.1

**Issue or Feature Request:**
I'd like to be able to connect to the admin container using EC2 Instance Connect.

jpculp:
#39 which adds support for EIC has been merged.

Thank you @samjo-nyang for the PR 🚀
and thank you @willthames for testing!

PR #39, opened 2021-06-28
Support EC2 instance connect

samjo-nyang:
**Issue number: #38 **

**Description of changes:**
- Install ec2-instance-connect package
- Update sshd_config file
- Make symbolic link of persistent ssh host key to /etc/ssh in the container
- Execute eic_harvest_hostkeys for one time

**Testing done:**
This patch is currently used in our internal Kubernetes cluster.

**Terms of contribution:**

### DIFF
--- a/start_admin_sshd.sh
+++ b/start_admin_sshd.sh
@@ -8,6 +8,8 @@ declare -r USER_DATA="${PERSISTENT_STORAGE_BASE_DIR}/user-data"
 
 declare -r LOCAL_USER="ec2-user"
 declare -r USER_SSH_DIR="/home/${LOCAL_USER}/.ssh"
+declare -r SSHD_CONFIG_DIR="/etc/ssh"
+declare -r SSHD_CONFIG_FILE="${SSHD_CONFIG_DIR}/sshd_config"
 
 # This is a counter used to verify at least
 # one of the methods below is available.
@@ -84,6 +86,25 @@ if trusted_user_ca_keys=$(get_user_data_keys "trusted-user-ca-keys") \
   ((++available_auth_methods))
 fi
 
+# Set additional configurations
+declare authorized_keys_command
+if authorized_keys_command=$(jq -e -r '.["ssh"]["authorized-keys-command"]?' "${USER_DATA}"); then
+  echo "AuthorizedKeysCommand ${authorized_keys_command}" >> "${SSHD_CONFIG_FILE}"
+fi
+
+declare authorized_keys_command_user
+if authorized_keys_command_user=$(jq -e -r '.["ssh"]["authorized-keys-command-user"]?' "${USER_DATA}"); then
+  echo "AuthorizedKeysCommandUser ${authorized_keys_command_user}" >> "${SSHD_CONFIG_FILE}"
+fi
+
+# Check the configurations are for EC2 instance connect
+declare -i use_eic=0
+if [[ $authorized_keys_command == /opt/aws/bin/eic_run_authorized_keys* ]] \
+&& [[ $authorized_keys_command_user == "ec2-instance-connect" ]]; then
+  use_eic=1
+  ((++available_auth_methods))
+fi
+
 chown -R "${LOCAL_USER}:" "${USER_SSH_DIR}"
 
 # If there were no successful auth methods, then users cannot authenticate
@@ -98,6 +119,7 @@ for key_alg in rsa ecdsa ed25519; do
   # If both of the keys exist, don't overwrite them
   if [[ -s "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key" ]] \
   && [[ -s "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub" ]]; then
+    ln -sf "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub" "${SSHD_CONFIG_DIR}/ssh_host_${key_alg}_key.pub"
     log "${key_alg} key already exists, will use existing key."
     continue
   fi
@@ -108,10 +130,19 @@ for key_alg in rsa ecdsa ed25519; do
   if ssh-keygen -t "${key_alg}" -f "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key" -q -N ""; then
     chmod 600 "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key"
     chmod 644 "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub"
+    ln -sf "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub" "${SSHD_CONFIG_DIR}/ssh_host_${key_alg}_key.pub"
   else
     log "Failure to generate host ${key_alg} ssh keys"
   fi
 done
+
+
+if [[ ${use_eic} == 1 ]] && [[ ! -f "${SSH_HOST_KEY_DIR}/harvest" ]]; then
+  if ! /opt/aws/bin/eic_harvest_hostkeys; then
+    log "Failure to harvest hostkeys for EIC"
+  fi
+  touch "${SSH_HOST_KEY_DIR}/harvest"
+fi
 
 install_proxy_profile
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ RUN test -n "$IMAGE_VERSION"
 LABEL "org.opencontainers.image.version"="$IMAGE_VERSION"
 
 RUN yum update -y \
-    && yum install -y openssh-server sudo util-linux procps-ng jq \
+    && yum install -y openssh-server sudo util-linux procps-ng jq openssl ec2-instance-connect \
     && yum clean all
 
 COPY --from=builder /opt/bash /opt/bin/
--- a/README.md
+++ b/README.md
@@ -34,6 +34,16 @@ To use custom public keys for `.ssh/authorized_keys` and/or custom CA keys for `
 }
 ```
 
+If you want to access to the admin container using [EC2 instance connect](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Connect-using-EC2-Instance-Connect.html), set `authorized-keys-command` and `authorized-keys-command-user` as follows:
+```
+{
+    "ssh": {
+      "authorized-keys-command": "/opt/aws/bin/eic_run_authorized_keys %u %f",
+      "authorized-keys-command-user": "ec2-instance-connect"
+    }
+}
+```
+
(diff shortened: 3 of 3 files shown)


start_admin_sshd.sh
```diff
@@ -98,6 +99,7 @@ for key_alg in rsa ecdsa ed25519; do
   # If both of the keys exist, don't overwrite them
   if [[ -s "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key" ]] \
   && [[ -s "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub" ]]; then
+    ln -sf "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub" "${SSHD_CONFIG_DIR}/ssh_host_${key_alg}_key.pub"
```
zmrow, line 122: I'm not sure I understand this - why are we symlinking keys there?

start_admin_sshd.sh
```diff
@@ -108,11 +110,19 @@ for key_alg in rsa ecdsa ed25519; do
   if ssh-keygen -t "${key_alg}" -f "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key" -q -N ""; then
     chmod 600 "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key"
     chmod 644 "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub"
+    ln -sf "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub" "${SSHD_CONFIG_DIR}/ssh_host_${key_alg}_key.pub"
   else
     log "Failure to generate host ${key_alg} ssh keys"
   fi
 done
 
+if [[ ! -f "${SSH_HOST_KEY_DIR}/harvest" ]]; then
```
zmrow: In the case a user doesn't want to use EC2 Instance Connect, we're still going to run these commands and log a failure. I'm wondering if there is a better trigger or something we can key off of in order to run/not run these.

sshd_config
```diff
@@ -17,3 +17,7 @@ Subsystem sftp	/usr/libexec/openssh/sftp-server
 
 # Configured by user data
 TrustedUserCAKeys /etc/ssh/trusted_user_ca_keys.pub
+
+# EC2 instance connect
+AuthorizedKeysCommand /opt/aws/bin/eic_run_authorized_keys %u %f
```
zmrow: From my perusal of the ssh docs, its seems to me that this will always run.  Am I reading that correctly?

zmrow (commented):
Thanks for the PR! :tada: 

Since additional setup is required to use EC2 Instance Connect, we'll also want to update the documentation to point at the AWS docs re: security group guidance and IAM permissions.  (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-connect-set-up.html)

sshd_config
samjo-nyang: https://github.com/aws/aws-ec2-instance-connect-config/blob/master/src/deb_systemd/ssh.service.d/ec2-instance-connect.conf#L3

Yes, this is the main entry point of ec2 instance connect.

https://github.com/aws/aws-ec2-instance-connect-config/blob/master/src/bin/eic_run_authorized_keys
It is a thin wrapper that calls eic_curl_authorized_keys

https://github.com/aws/aws-ec2-instance-connect-config/blob/master/src/bin/eic_curl_authorized_keys
The eic_curl_authorized_keys fetches valid key signature from the ec2 instance connect metadata servers, and validate ssh requests. Therefore, it should be always run to use ec2-instance-connect.

start_admin_sshd.sh
samjo-nyang: Umm... are there any methods to fetch dynamic configuration from ec2 metadata or bottlerocket host such as `enable_eic: 1`?

start_admin_sshd.sh
samjo-nyang: Actually, it is a kind of trick - as you know, SSH_HOST_KEY_DIR is a permanent while SSHD_CONFIG_DIR (/etc/ssh) is not. Therefore, to make SSHD_CONFIG_DIR persistent across admin container reboot, I just make symbolic links.

The reason why SSHD_CONFIG_DIR should be persistent, is that the eic_harvest_hostkeys scripts refers /etc/ssh to read host keys, and the path is hard-coded on the script.
https://github.com/aws/aws-ec2-instance-connect-config/blob/master/src/bin/eic_harvest_hostkeys#L82

jpculp: I wonder if it would be beneficial to make the whole `/etc/ssh/` directory a symlink to `/.bottlerocket/host-containers/current/etc/ssh/` as opposed to individual files.

start_admin_sshd.sh
jpculp: Currently we pass things to the admin container via the user-data json. At the moment it's only used for ssh keys, but perhaps we could also pass an enable-eic key/value pair. `imdsclient` and`shibaken` could potentially fetch it from IMDS (if it's there). Anyone have any thoughts on that?

start_admin_sshd.sh
samjo-nyang: It sounds good, but I have a question: sshd_config file is also in the /etc/ssh/ directory, and the file copy done at the Dockerfile build time (https://github.com/bottlerocket-os/bottlerocket-admin-container/blob/develop/Dockerfile#L62)
Should I change the Dockerfile to copy the sshd_config to `/etc/ssh-template/` and `ln -s ... && mv /etc/ssh-template/sshd_config /etc/ssh/sshd_config` in this start script?

start_admin_sshd.sh
samjo-nyang:
```
declare -i eic_enabled=0
if ! eic_enabled=$(jq -e -r '.["enable_eic"]?' "${USER_DATA}"); then
  eic_enabled=1
fi
```
I just add this block to fetch "enable_eic" from userdata and make its default as 1 (if not exists); and execute the harvest script if and only if enable_eic=1.

start_admin_sshd.sh
jpculp: I tried it out yesterday and I'm not sure it's worth the hassle. I'm fine with your current symlink approach 😄

jpculp (commented):
Sorry for the delayed response. After thinking about it a bit more, it might make more sense to have the userdata structure look more like:
```
{
   "ssh":{
      "authorized-keys":[
         "ssh-rsa EXAMPLEAUTHORIZEDPUBLICKEYHERE my-key-pair"
      ],
      "trusted-user-ca-keys":[
         "ssh-rsa EXAMPLETRUSTEDCAPUBLICKEYHERE authority@ssh-ca.example.com"
      ],
      "authorized-keys-command": "...",
      "authorized-keys-command-user": "..."
}
```

You could parse the authorized keys command and user and output them to their own config under `/etc/ssh/sshd_config.d/`. If those values are for **EC2 Instance Connect** you could conditionally trigger the symlink and key harvest logic you have in the current PR. We are going to double-check on our end to see if the key harvest logic is required or whether it is safe to omit. Is this something you would be open to implementing? Or would you be more comfortable with us working on it?

jpculp:
I've verified that `eic_harvest_hostkeys` is used for establishing trust, so we should not omit it.

samjo-nyang:
@jpculp Thanks for your review!
I re-implemented the logic using "authorized-keys-command" and "authorized-keys-command-user" userdata. Is this fine?
Also, I force-pushed this branch to squash commits and fix author info.

jpculp (changes requested):
Thank you for the quick update @samjo-nyang! I think this is very close.

I'm afraid I led you astray by suggesting the switch to a separate config file. After some testing on my end, it seems that the OpenSSH in the admin container does not yet support the `Include` directive in the `sshd_config` file itself. It might be better to just append to `sshd_config` directly. You could have a variable like **SSHD_CONFIG_FILE** variable pointing to  `"${SSHD_CONFIG_DIR}/sshd_config"`.

samjo-nyang:
@jpculp Updated! (and also force-pushed)

willthames:
I currently get
```
[  321.634455] host-ctr[7917]: Failed to configure ssh authentication with user-data: {"ssh":{"authorized-keys-command":"/opt/aws/bin/eic_run_authorized_keys %u %f","authorized-keys-command-user":"ec2-instance-connect"}}
[  321.648738] host-ctr[7917]: rsa key already exists, will use existing key.
[  321.651485] host-ctr[7917]: ecdsa key already exists, will use existing key.
[  321.654770] host-ctr[7917]: ed25519 key already exists, will use existing key.
[  321.665726] host-ctr[7917]: /etc/ssh/sshd_config: line 22: Bad configuration option: Include
[  321.668964] host-ctr[7917]: /etc/ssh/sshd_config: terminating, 1 bad configuration options
[  321.708568] host-ctr[7917]: time="2021-07-09T02:56:48Z" level=fatal msg="Container admin exited with non-zero status"
```

I understand the Include problem is mentioned and fixed above, but the userdata problem would remain.

My reading is that this happens because setting authorized-keys-command and authorized-keys-command-user does not increment available_auth_methods

samjo-nyang:
@willthames I just re-ordered the statements and increase available_auth_methods when use_eic=1. It should work now.

willthames:
Thanks @samjo-nyang  - I'll give it a go!

willthames:
This works great for me now.

start_admin_sshd.sh
```diff
@@ -98,6 +119,9 @@ for key_alg in rsa ecdsa ed25519; do
   # If both of the keys exist, don't overwrite them
   if [[ -s "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key" ]] \
   && [[ -s "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub" ]]; then
+    if [[ ${use_eic} == 1 ]]; then
```
jpculp: Nit. It's safe to remove this conditional and just symlink regardless, but I'm also fine with it as is.

start_admin_sshd.sh
```diff
@@ -108,11 +132,22 @@ for key_alg in rsa ecdsa ed25519; do
   if ssh-keygen -t "${key_alg}" -f "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key" -q -N ""; then
     chmod 600 "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key"
     chmod 644 "${SSH_HOST_KEY_DIR}/ssh_host_${key_alg}_key.pub"
+    if [[ ${use_eic} == 1 ]]; then
```
jpculp: Same as above ↑

jpculp (approved):
This looks great to me!

Also, thank you for testing this out @willthames!

samjo-nyang:
Force-pushed (remove use_eic=1 condition and symlink the host keys for all cases)

bcressey (approved):

webern (approved):

Issue #41, opened 2021-07-16
Improve sshd configuration


willthames:
**Image I'm using:**

Custom image based on pr #39 but that should be pretty much latest now.

**Issue or Feature Request:**

I would like either a stricter sshd config or the ability to configure a stricter sshd config.

e.g. 

```
Ciphers chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com
PermitRootLogin no
```

I can see two ways to achieve that:
* hardcode it for everyone in sshd_config
* add a `custom-config` (or better name) to user data e.g.
  ```
  {
    "ssh":{
       "custom-config": {
          "PermitRootLogin": "no",
          "Ciphers": "chacha20-poly1305@openssh.com,aes128-ctr,aes192-ctr,aes256-ctr,aes128-gcm@openssh.com,aes256-gcm@openssh.com"
       }
    }
  }
  ```

I am willing to implement this but just want to check what your preferences are first.

jpculp:
@willthames, thank you for opening this enhancement request! After some discussion with the team, we agree that this would be a great change. We would prefer hardcoding the `PermitRootLogin no` directly to the `sshd_config` above the `# Configured by user data` line, and setting the ciphers via user-data like so:
```
{
    "ssh": {
        "authorized-keys-command": "/opt/aws/bin/eic_run_authorized_keys %u %f",
        "authorized-keys-command-user": "ec2-instance-connect",
        "ciphers": [
            "chacha20-poly1305@openssh.com",
            "aes128-ctr",
            "aes192-ctr",
            "aes256-ctr",
            "aes128-gcm@openssh.com",
            "aes256-gcm@openssh.com"
        ]
    }
}
```
Is this something you would still be willing to implement or would you rather us work on it?

willthames:
@jpculp - I'm happy with that approach - I'll hopefully submit a PR shortly

willthames:
Closed by #42

PR #42, opened 2021-07-19
Improve sshd configuration

willthames:
**Issue number:**

#41

**Description of changes:**

Disable root login

Allow custom SSH cipher list to be set

**Testing done:**

Created a container from the image and ensured that I can log in and that /etc/ssh/sshd_config looks correct

**Terms of contribution:**

### DIFF
--- a/start_admin_sshd.sh
+++ b/start_admin_sshd.sh
@@ -97,6 +97,11 @@ if authorized_keys_command_user=$(jq -e -r '.["ssh"]["authorized-keys-command-us
   echo "AuthorizedKeysCommandUser ${authorized_keys_command_user}" >> "${SSHD_CONFIG_FILE}"
 fi
 
+# Populate ciphers with all the ciphers found in user-data
+if ciphers=$(jq -c '.["ssh"]["ciphers"]?' "${USER_DATA}" | tr -d '[]"'); then
+  echo "Ciphers ${ciphers}" >> "${SSHD_CONFIG_FILE}"
+fi
+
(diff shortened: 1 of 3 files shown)


start_admin_sshd.sh
```diff
@@ -97,6 +97,11 @@ if authorized_keys_command_user=$(jq -e -r '.["ssh"]["authorized-keys-command-us
   echo "AuthorizedKeysCommandUser ${authorized_keys_command_user}" >> "${SSHD_CONFIG_FILE}"
 fi
 
+# Populate ciphers with all the ciphers found in user-data
+if ciphers=$(jq -c '.["ssh"]["ciphers"]?' "${USER_DATA}" | tr -d '[]"'); then
```
tjkirch:
```suggestion
if ciphers=$(jq -r -e -c '.["ssh"]["ciphers"]? | join(",")' "${USER_DATA}" 2>/dev/null); then
```
We should use `-e` so that the `if` can fail if jq does, otherwise we'd print "Ciphers null".  `-r` means we won't have to strip off the quotes.  `join` can replace the `tr`.  (`tr` was changing the return code to 0 anyway, which would have necessitated `set -o pipefail`, but this seems simpler.)

tjkirch (commented):
Thanks again!

willthames:
@tjkirch I've retested with your jq improvements - works well for me, thanks!

tjkirch (approved):

arnaldo2792 (approved):

jpculp (approved):
🚀

Issue #31, opened 2021-04-30
Use a different username than `ec2-user`


zmrow:
**Image I'm using:**
v0.7.0

**Issue or Feature Request:**
When running Bottlerocket outside of EC2 it's odd to login to the admin container with `ec2-user@...`.

We could use a generic user name instead, something like `admin-user`.  The change may be jarring for folks currently running on EC2, so potentially we could have both for a short time?

jpculp:
Merged code to specify user in host-container user-data.

PR #45, opened 2021-08-04
Support custom username

jpculp:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket-admin-container/issues/31

**Description of changes:**

By default, the admin container's local user is `ec2-user`.
The first commit allows for the local user to be named something else.

The second commit cleans up the `README.md` by normalizing spacing of the examples.

**Testing done:**

- [x] No custom user-data.
- [x] User-data with custom user: `"user-data": "bottlerocket"`
- [x] User-data with malformed data: `"user-data": {]`
- [x] User-data with an empty string: `"user-data": ""`

**Terms of contribution:**

### DIFF
--- a/start_admin_sshd.sh
+++ b/start_admin_sshd.sh
@@ -2,11 +2,21 @@
 
 set -e
 
+log() {
+  echo "$*" >&2
+}
+
 declare -r PERSISTENT_STORAGE_BASE_DIR="/.bottlerocket/host-containers/current"
 declare -r SSH_HOST_KEY_DIR="${PERSISTENT_STORAGE_BASE_DIR}/etc/ssh"
 declare -r USER_DATA="${PERSISTENT_STORAGE_BASE_DIR}/user-data"
 
-declare -r LOCAL_USER="ec2-user"
+# Fetch user from user-data json (if any). Default to 'ec2-user' if null or invalid.
+if ! LOCAL_USER=$(jq -e -r '.["user"] // "ec2-user"' "${USER_DATA}" 2>/dev/null) \
+|| [[ ! "${LOCAL_USER}" =~ ^[a-z_][a-z0-9_-]{0,31}$ ]]; then
+  log "Failed to set user from user-data. Proceeding with 'ec2-user'."
+  LOCAL_USER="ec2-user"
+fi
+
 declare -r USER_SSH_DIR="/home/${LOCAL_USER}/.ssh"
 declare -r SSHD_CONFIG_DIR="/etc/ssh"
 declare -r SSHD_CONFIG_FILE="${SSHD_CONFIG_DIR}/sshd_config"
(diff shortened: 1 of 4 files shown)


start_admin_sshd.sh
```diff
@@ -63,6 +68,10 @@ EOF
   chmod 644 "${proxy_profile}"
 }
 
+# Create local user
+echo "${LOCAL_USER} ALL=(ALL) NOPASSWD: ALL" > "/etc/sudoers.d/${LOCAL_USER}"
+chmod 440 "/etc/sudoers.d/${LOCAL_USER}"
+useradd -m -G users,api "${LOCAL_USER}"
```
jpculp, line 75: In the original Dockerfile the `useradd` came after creating the sudoers file. Does it make more sense to move it to the top here?

start_admin_sshd.sh
```diff
@@ -2,11 +2,20 @@
...
 declare -r PERSISTENT_STORAGE_BASE_DIR="/.bottlerocket/host-containers/current"
 declare -r SSH_HOST_KEY_DIR="${PERSISTENT_STORAGE_BASE_DIR}/etc/ssh"
 declare -r USER_DATA="${PERSISTENT_STORAGE_BASE_DIR}/user-data"
 
-declare -r LOCAL_USER="ec2-user"
+# Fetch custom user from user-data json (if any). Default to ec2-user if null or failed.
+if ! LOCAL_USER=$(jq -e -r '.["user"] // "ec2-user"' "${USER_DATA}" 2>/dev/null); then
+  log "Failed to set user with user-data. Proceeding with 'ec2-user'."
+  declare -r LOCAL_USER="ec2-user"
+fi
```
tjkirch:
```suggestion
LOCAL_USER="$(jq -e -r '.["user"] // "ec2-user"' "${USER_DATA}" 2>/dev/null)"
if [ "${?}" -ne 0 ] || [[ ! ${LOCAL_USER} =~ ^[a-z_][a-z0-9_-]{0,31}$ ]]; then
  log "Failed to set user from user-data. Proceeding with 'ec2-user'."
  LOCAL_USER="ec2-user"
fi
```
The `// "ec2-user"` means that the only time we hit the conditional is from bad JSON syntax.  `//` is nice to prevent outputting false/null, but unfortunately it doesn't catch empty strings, so user data with `{"user": ""}` would set LOCAL_USER="" and cause later commands to fail.  So, I think we need the extra check.  And if we're adding the check, we might as well do a better one than -z, since invalid characters would also cause later commands to fail.  (I also suggested removing the `declare -r` because having the success case be writable and the failure case not seems odd?)

start_admin_sshd.sh
tjkirch: I'm not sure it matters; the sudoers file isn't used until the user tries to sudo.  It makes sense to me to keep the commands together like you have...

jpculp:
- Tweaked the logic to verify the validity of the user specified in user-data after the jq.
- Added test case for user being set to an empty string.

tjkirch (approved):

jpculp:
I improved the logic a little to be more condensed, but also factor in the scenario where jq returns an error that somehow passes the regex.

zmrow (approved):
LGTM 

:joystick:

tjkirch (approved):

jpculp:
Made a small cosmetic change for consistency and retested.

tjkirch (approved):

zmrow (approved):

PR #50, opened 2021-11-13
sheltie: add ability to run single commands

jpculp:
**Description of changes:**

Rather than entering a full `bash` shell, you can pass a command as an argument.

Example: `sudo sheltie cat /etc/os-release`.

**Testing done:**

- Running `sudo sheltie` put me in an interactive `bash` shell.
- Running `sudo sheltie cat /etc/os-release` simply printed my Bottlerocket release info.

**Terms of contribution:**

### DIFF
--- a/sheltie
+++ b/sheltie
@@ -18,4 +18,8 @@ ROOT_FS_PATH="/proc/${PPID}/root"
 BASH_PATH="/opt/bin/bash"
 
 # Start the root shell on the Bottlerocket host
-exec nsenter -t 1 -a "${ROOT_FS_PATH}${BASH_PATH}"
+if [[ ${#@} -ne 0 ]]; then
+  exec nsenter -t 1 -a -- "$@"
+else
+  exec nsenter -t 1 -a "${ROOT_FS_PATH}${BASH_PATH}"
+fi


tjkirch (approved):
This gives us the flexibility to add sheltie-specific arguments later, too, if we need.

sheltie
```diff
@@ -18,4 +18,8 @@ ROOT_FS_PATH="/proc/${PPID}/root"
 BASH_PATH="/opt/bin/bash"
 
 # Start the root shell on the Bottlerocket host
-exec nsenter -t 1 -a "${ROOT_FS_PATH}${BASH_PATH}"
+if [[ -n "$*" ]]; then
```
tjkirch:
```suggestion
if [ "${#@}" -ne 0 ]; then
```
nit: better to check the size directly rather than the length of its string expansion.  (If the user passes an empty string, for some reason, it *should* fail, we shouldn't default to bash if, for example, they intended to run a command from a variable but it wasn't set and resulted in an empty string.)

jpculp: From what I could tell in my test container, the above change didn't alter the behavior. I'd prefer to keep as is. I think if we wanted to fail if someone intended to pass something, but didn't, we would need to add something like `-c` so that if the flag is there, but nothing is after it, an error can be printed.

samuelkarp:
```
$ cat args.sh
[[ -n "$*" ]] && echo "Found non-zero string of args"
[[ "${#@}" -ne 0 ]] && echo "Found non-zero array of args"
$ ./args.sh
$ ./args.sh one
Found non-zero string of args
Found non-zero array of args
$ ./args.sh ""
Found non-zero array of args
```

That should illustrate the difference in behavior for you.

jpculp: Ah, when I tested passing with `BOB="" sudo sheltie $BOB` I didn't see a difference in behavior, but when I passed `BOB="" sudo sheltie "$BOB"` then I do get `nsenter: failed to execute : No such file or directory`. I'm not super keen on different behavior based on whether an empty variable is quoted or not, but I'm willing to adjust it if you still think it is necessary.

samuelkarp: I agree with @tjkirch that it should fail on empty strings, so I'd like to see this changed.

jpculp:
- Replaced `"$*"` with `${#@}` per @tjkirch @samuelkarp

tjkirch (approved):

zmrow (approved):

samuelkarp (approved):

PR #51, opened 2021-12-01
Run yum update during Docker image build

cbgbt:
Description of changes:
This performs a yum update of the underlying Amazon Linux container prior to building the control container.

Testing done:

Confirmed that the build successfully performs a yum update prior to building the rest of the image.
Terms of contribution:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 ARG DOCKER_ARCH
 FROM $DOCKER_ARCH/amazonlinux:2 as builder
+RUN yum update -y
 RUN yum group install -y "Development Tools"
 RUN yum install -y glibc-static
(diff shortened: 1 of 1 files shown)


webern (approved):

Dockerfile
```diff
@@ -1,5 +1,6 @@
 ARG DOCKER_ARCH
 FROM $DOCKER_ARCH/amazonlinux:2 as builder
+RUN yum update -y
 RUN yum group install -y "Development Tools"
 RUN yum install -y glibc-static
```
bcressey, line 5: It looks like this layer is only used for building `sheltie`, and we actually already do a `yum update -y` for the final layer.

PR #59, opened 2022-04-28
Serial console access support

etungsten:
**Issue number:**
N/A

**Description of changes:**

```
    Serial console support

    This adds getty services for each serial console device on the host so
    users can get to the admin container via serial console connections.

    As part of this change, the sshd process is now managed by a
    user-managed instance of systemd running in the admin container.

    All the admin container user-managed services are "WantedBy" the
    admin.target.

    Users can specify a password hash for the console login in the admin
    host container's user-data.

```

**Testing done:**
With no admin user-data specified. I was able to connect to the admin container via a serial console connection to my EC2 instance.
`sudo sheltie` works,
`apiclient` works.
I can still SSH to the admin container.
```
$ ssh -i ~/pem/etung.pem i-01b9b773cfed7c6c9.port0@serial-console.ec2-instance-connect.us-west-2.aws
[   30.635206] host-ctr[1605]: Created symlink /root/.config/systemd/user/admin.target.wants/getty@tty0.service, pointing to /etc/systemd/user/getty@.service.
[   30.694509] host-ctr[1605]: Created symlink /root/.config/systemd/user/admin.target.wants/serial-getty@ttyS0.service, pointing to /etc/systemd/user/serial-getty@.service.
[   30.738476] host-ctr[1605]: Created symlink /root/.config/systemd/user/admin.target.wants/sshd.service, pointing to /etc/systemd/user/sshd.service.
[   30.776554] host-ctr[1605]: Startup finished in 66ms.

Last login: Wed Apr 27 23:59:41 on /.bottlerocket/rootfs/dev/tty0
          Welcome to Bottlerocket's admin container!
    ╱╲
   ╱┄┄╲   This container provides access to the Bottlerocket host
   │▗▖│   filesystems (see /.bottlerocket/rootfs) and contains common
  ╱│  │╲  tools for inspection and troubleshooting.  It is based on
  │╰╮╭╯│  Amazon Linux 2, and most things are in the same places you
    ╹╹    would find them on an AL2 host.

To permit more intrusive troubleshooting, including actions that mutate the
running state of the Bottlerocket host, we provide a tool called "sheltie"
(`sudo sheltie`).  When run, this tool drops you into a root shell in the
Bottlerocket host's root filesystem.
[ec2-user@admin]$ env
HOSTNAME=
TERM=xterm-256color
SHELL=/bin/bash
HISTSIZE=1000
NO_PROXY=localhost,127.0.0.1,....yl4.us-west-2.eks.amazonaws.com,.cluster.local
USER=ec2-user
LS_COLORS=....
MAIL=/var/spool/mail/ec2-user
PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/ec2-user/.local/bin:/home/ec2-user/bin
PWD=/home/ec2-user
HISTCONTROL=ignoredups
SHLVL=1
HOME=/home/ec2-user
no_proxy=localhost,127.0.0.1,....yl4.us-west-2.eks.amazonaws.com,.cluster.local
LOGNAME=ec2-user
_=/usr/bin/env
[ec2-user@admin]$ sudo sheltie
bash-5.1# exit
[ec2-user@admin]$ apiclient get settings.host-containers.admin
{
  "settings": {
    "host-containers": {
      "admin": {
        "enabled": true,
        "source": "...",
        "superpowered": true,
        "user-data": "...."
      }
    }
  }
}
```

I then set up a new bottlerocket host with admin container user-data containing a password-hash for the console login for my custom user:
Userdata:
```
{
  "user": "erikson",
  "password-hash": "$6$rounds=1000$NER...",
  "ssh": {
    "authorized-keys...",
  }
}
```

Then trying to login via serial console:
```
ip-192-168-2-248 login: erikson
Password:
Last failed login: Thu Apr 28 00:33:17 UTC 2022 on /.bottlerocket/rootfs/dev/ttyS0
There were 2 failed login attempts since the last successful login.
          Welcome to Bottlerocket's admin container!
    ╱╲
   ╱┄┄╲   This container provides access to the Bottlerocket host
   │▗▖│   filesystems (see /.bottlerocket/rootfs) and contains common
  ╱│  │╲  tools for inspection and troubleshooting.  It is based on
  │╰╮╭╯│  Amazon Linux 2, and most things are in the same places you
    ╹╹    would find them on an AL2 host.

To permit more intrusive troubleshooting, including actions that mutate the
running state of the Bottlerocket host, we provide a tool called "sheltie"
(`sudo sheltie`).  When run, this tool drops you into a root shell in the
Bottlerocket host's root filesystem.
[erikson@admin]$ 
```

- [x] test on bare metal server bottlerocket host

**Terms of contribution:**

### DIFF
--- a/start_admin.sh
+++ b/start_admin.sh
@@ -10,6 +10,11 @@ declare -r PERSISTENT_STORAGE_BASE_DIR="/.bottlerocket/host-containers/current"
 declare -r SSH_HOST_KEY_DIR="${PERSISTENT_STORAGE_BASE_DIR}/etc/ssh"
 declare -r USER_DATA="${PERSISTENT_STORAGE_BASE_DIR}/user-data"
 
+if [ ! -s "${USER_DATA}" ]; then
+  log "Admin host-container user-data is empty, going to sleep forever"
+  exec sleep infinity
+fi
+
 # Fetch user from user-data json (if any). Default to 'ec2-user' if null or invalid.
 if ! LOCAL_USER=$(jq -e -r '.["user"] // "ec2-user"' "${USER_DATA}" 2>/dev/null) \
 || [[ ! "${LOCAL_USER}" =~ ^[a-z_][a-z0-9_-]{0,31}$ ]]; then
--- a/start_admin.sh
+++ b/start_admin.sh
@@ -111,7 +111,11 @@ enable_systemd_services() {
 # Create local user
 echo "${LOCAL_USER} ALL=(ALL) NOPASSWD: ALL" > "/etc/sudoers.d/${LOCAL_USER}"
 chmod 440 "/etc/sudoers.d/${LOCAL_USER}"
-useradd -m -G users,api "${LOCAL_USER}"
+# Skip user creation if the user already exists
+if ! id -u "${LOCAL_USER}" &>/dev/null; then
+  useradd -m "${LOCAL_USER}"
+fi
+usermod -G users,api "${LOCAL_USER}"
 if [[ -z "${PASSWORD_HASH}" ]]; then
   usermod -L "${LOCAL_USER}"
 else
--- a/start_admin.sh
+++ b/start_admin.sh
@@ -17,13 +17,18 @@ if ! LOCAL_USER=$(jq -e -r '.["user"] // "ec2-user"' "${USER_DATA}" 2>/dev/null)
   LOCAL_USER="ec2-user"
 fi
 
+# Fetch password-hash for serial console access.
+if ! PASSWORD_HASH=$(jq -r '.["password-hash"] // ""' "${USER_DATA}" 2>/dev/null); then
+  PASSWORD_HASH=""
+fi
+
 declare -r USER_SSH_DIR="/home/${LOCAL_USER}/.ssh"
 declare -r SSHD_CONFIG_DIR="/etc/ssh"
 declare -r SSHD_CONFIG_FILE="${SSHD_CONFIG_DIR}/sshd_config"
 
 # This is a counter used to verify at least
 # one of the methods below is available.
-declare -i available_auth_methods=0
+declare -i available_ssh_methods=0
 
 get_user_data_keys() {
     # Extract the keys from user-data json
@@ -69,34 +74,73 @@ EOF
   chmod 644 "${proxy_profile}"
 }
 
+# Set up agetty services for serial console access and the sshd daemon service
+enable_systemd_services() {
+  # Grab `console=` parameters from the kernel command line.
+  CONSOLES=()
+  for opt in $(cat /proc/cmdline) ; do
+     optarg="$(expr "${opt}" : '[^=]*=\(.*\)' 2>/dev/null ||:)"
+     optarg="${optarg%\"}"
+     optarg="${optarg#\"}"
+     case "${opt}" in
+        console=*) CONSOLES+=("${optarg%,*}") ;;
+     esac
+  done
+
+  HOST_DEVTMPFS="/.bottlerocket/rootfs/dev"
+  for console in "${CONSOLES[@]}" ; do
+     # Skip devices that don't exist.
+     [ -c "${HOST_DEVTMPFS}/${console}" ] || continue
+
+     # Otherwise instantiate a service from the template unit. This is normally
+     # done by `systemd-getty-generator`, but that skips over ordinary devices
+     # when run inside a container.
+     case "${console}" in
+        ttyS*|ttyAMA*|ttyUSB*)
+           systemctl --user enable "serial-getty@${console}.service"
+           ;;
+        tty*)
+           systemctl --user enable "getty@${console}.service"
+           ;;
+     esac
+  done
+  # Enable the SSH daemon service unit so we can run it in the background
+  systemctl --user enable "sshd.service"
+}
+
 # Create local user
 echo "${LOCAL_USER} ALL=(ALL) NOPASSWD: ALL" > "/etc/sudoers.d/${LOCAL_USER}"
 chmod 440 "/etc/sudoers.d/${LOCAL_USER}"
 useradd -m -G users,api "${LOCAL_USER}"
+if [[ -z "${PASSWORD_HASH}" ]]; then
+  usermod -L "${LOCAL_USER}"
+else
+  usermod -p "${PASSWORD_HASH}" "${LOCAL_USER}"
+fi
 mkdir -p "${USER_SSH_DIR}"
 chmod 700 "${USER_SSH_DIR}"
 
-# Populate authorized_keys with all the authorized keys found in user-data
+# Populate SSH authorized_keys with all the authorized keys found in user-data
 if authorized_keys=$(get_user_data_keys "authorized-keys") \
 || authorized_keys=$(get_user_data_keys "authorized_keys"); then
   ssh_authorized_keys="${USER_SSH_DIR}/authorized_keys"
   touch "${ssh_authorized_keys}"
   chmod 600 "${ssh_authorized_keys}"
   echo "${authorized_keys}" > "${ssh_authorized_keys}"
-  ((++available_auth_methods))
+  ((++available_ssh_methods))
 fi
 
-# Populate trusted_user_ca_keys with all the trusted ca keys found in user-data
+# Populate SSH trusted_user_ca_keys with all the trusted ca keys found in user-data
 if trusted_user_ca_keys=$(get_user_data_keys "trusted-user-ca-keys") \
 || trusted_user_ca_keys=$(get_user_data_keys "trusted_user_ca_keys"); then
   ssh_trusted_user_ca_keys="/etc/ssh/trusted_user_ca_keys.pub"
   touch "${ssh_trusted_user_ca_keys}"
   chmod 600 "${ssh_trusted_user_ca_keys}"
   echo "${trusted_user_ca_keys}" > "${ssh_trusted_user_ca_keys}"
-  ((++available_auth_methods))
+  ((++available_ssh_methods))
 fi
 
-# Set additional configurations
+# Set additional SSH configurations
 declare authorized_keys_command
 if authorized_keys_command=$(jq -e -r '.["ssh"]["authorized-keys-command"]?' "${USER_DATA}"); then
   echo "AuthorizedKeysCommand ${authorized_keys_command}" >> "${SSHD_CONFIG_FILE}"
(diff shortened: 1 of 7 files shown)


start_admin.sh
```diff
@@ -127,10 +171,10 @@ fi
 
 chown -R "${LOCAL_USER}:" "${USER_SSH_DIR}"
 
-# If there were no successful auth methods, then users cannot authenticate
+# If there were no available SSH auth methods, then users cannot connect via the SSH daemon
 if [[ "${available_auth_methods}" -eq 0 ]]; then
```
jpculp:
```suggestion
if [[ "${available_ssh_methods}" -eq 0 ]]; then
```
Since we are only counting ssh authentication methods, it would probably be best to rename this to reduce confusion.

start_admin.sh
```diff
@@ -69,14 +76,51 @@ EOF
...
+  done
+  # Enable the SSH daemon service unit so we can run it in the background
+  systemctl --user enable "sshd.service"
+}
+
 # Create local user
 echo "${LOCAL_USER} ALL=(ALL) NOPASSWD: ALL" > "/etc/sudoers.d/${LOCAL_USER}"
 chmod 440 "/etc/sudoers.d/${LOCAL_USER}"
 useradd -m -G users,api "${LOCAL_USER}"
+if [ -z "${PASSWORD_HASH}" ]; then
```
jpculp: nit. but it might be nice for `[[..]]` for consistency in the script.

jpculp (commented):
LGTM, pending metal testing.

start_admin.sh
```diff
@@ -69,14 +76,51 @@ EOF
...
+  passwd -d "${LOCAL_USER}"
```
etungsten: If the "password-hash" is not specified in the admin container user-data, then this deletes the password for the user to allow passwordless login. Not sure if we want that or just disable password logins to not allow console logins.

bcressey: We want to disable password logins in that case (`usermod -L "${LOCAL_USER}"`, assuming that still allows ssh) to avoid exposing a password-less serial console on upgrade. It's not a given that someone with serial console access should have the ability to log into a privileged account on the host.

units/getty@.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Getty on %I
+DefaultDependencies=no
+
+[Service]
+Environment=TERM=xterm-256color
+EnvironmentFile=/etc/getty/username.env
+ExecStart=-/sbin/agetty -o '-- ${USERNAME}' --noclear --skip-login - ${TERM}
+Type=idle
```
bcressey, line 8: One change worth experimenting with is whether Bottlerocket itself should run host container units with `Type=idle`, to ensure that we don't suppress normal boot output as soon as the admin container starts.

units/getty@.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Getty on %I
+DefaultDependencies=no
+
+[Service]
+Environment=TERM=xterm-256color
+EnvironmentFile=/etc/getty/username.env
+ExecStart=-/sbin/agetty -o '-- ${USERNAME}' --noclear --skip-login - ${TERM}
```
bcressey: Does using `${USERNAME}` rather than `\u` mean that no matter what username is entered, we always send the user-data-supplied username to `login`? That's arguably a minor loss to security, since they wouldn't have to guess the valid username.

bcressey: We can potentially also keep the `-p` flag if we use an `UnsetEnvironment=<blah>` directive below to remove problematic environment variables.

That would allow us to pass through variables like `NO_PROXY` that are set in the ssh case, which I'd guess might be discarded otherwise.

units/sshd.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=OpenSSH server daemon
+Documentation=man:sshd(8) man:sshd_config(5)
```
bcressey: We can probably delete this documentation line since we don't have `man` or man pages in the container.

units/sshd.service
```diff
@@ -0,0 +1,15 @@
...
+
+[Service]
+Type=notify
+EnvironmentFile=/etc/sysconfig/sshd
+ExecStart=/usr/sbin/sshd -e -D $OPTIONS
+ExecReload=/bin/kill -HUP $MAINPID
+KillMode=process
+Restart=on-failure
+RestartSec=42s
```
bcressey, line 11: 🤣

bcressey (commented):
Nice!

units/getty@.service
etungsten: It's currently not suppressing normal boot output even after the admin container starts. Not sure if that's desirable though:
```
[erikson@admin]$ [  119.581249] IPv6: ADDRCONF(NETDEV_CHANGE): eni43a75d0d3cf: link becomes ready
[  119.586160] IPv6: ADDRCONF(NETDEV_CHANGE): eth0: link becomes ready
[  120.203507] IPv6: ADDRCONF(NETDEV_CHANGE): eth0: link becomes ready
...
```

units/getty@.service
etungsten: I tried `UnsetEnvironment` and it didn't unset `HOME=/root`.

Interestingly enough, without `-p` the proxy env vars still show up fine. I assume because `login` doesn't try to set those but wants to set `HOME`

```
[erikson@admin]$ env
HOSTNAME=
TERM=xterm-256color
SHELL=/bin/bash
HISTSIZE=1000
NO_PROXY=localhost,127.0.0.1,191bc2a1c5d1e9e863c973a7bdfb88d0.yl4.us-west-2.eks.amazonaws.com,.cluster.local
USER=erikson
...
MAIL=/var/spool/mail/erikson
PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/erikson/.local/bin:/home/erikson/bin
PWD=/home/erikson
HISTCONTROL=ignoredups
SHLVL=1
HOME=/home/erikson
no_proxy=localhost,127.0.0.1,191bc2a1c5d1e9e863c973a7bdfb88d0.yl4.us-west-2.eks.amazonaws.com,.cluster.local
LOGNAME=erikson
_=/usr/bin/env

```

etungsten:
Push above addresses comments.

- Removed `--skip-login` from `getty` services
- Disable password login if `password-hash` is not specified in the admin container's host-container user-data.

start_admin.sh
```diff
@@ -17,13 +17,20 @@ if ! LOCAL_USER=$(jq -e -r '.["user"] // "ec2-user"' "${USER_DATA}" 2>/dev/null)
   LOCAL_USER="ec2-user"
 fi
 
+# Write out environment files for getty services
+mkdir -p /etc/getty
+echo "USERNAME=${LOCAL_USER}" > "/etc/getty/username.env"
```
jpculp: Is this still necessary with the latest change?

start_admin.sh
```diff
@@ -122,15 +166,15 @@ declare -i use_eic=0
...
 chown -R "${LOCAL_USER}:" "${USER_SSH_DIR}"
 
-# If there were no successful auth methods, then users cannot authenticate
-if [[ "${available_auth_methods}" -eq 0 ]]; then
+# If there were no available SSH auth methods, then users cannot connect via the SSH daemon
+if [[ "${available_ssh_methods}" -eq 0 ]]; then
   user_data_condensed=$(jq -e -c . "${USER_DATA}" 2>/dev/null || cat "${USER_DATA}")
-  log "Failed to configure ssh authentication with user-data: ${user_data_condensed}"
+  log "No SSH authentication methods available in admin container user-data: ${user_data_condensed}"
 fi
```
jpculp: Do we still want to print out user data in the event that SSH authentication is not configured now that we're adding serial console access?

units/getty@.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Getty on %I
+DefaultDependencies=no
+
+[Service]
+Environment=TERM=xterm-256color
+UnsetEnvironment=HOME=/root
```
bcressey: Did this end up having some effect?

bcressey (approved):

units/getty@.service
bcressey: I'm wondering whether we still see output from e.g. systemd services starting after agetty starts, or if it just ends at the login prompt except for the occasional kernel message.

It might be hard to observe since the admin container takes a while to pull and start, but you could add a sleep to `kubelet` to see if we get both Starting + Started messages.

units/getty@.service
etungsten: Ah, it didn't. I'll go ahead and remove it.

units/getty@.service
etungsten: I put a sleep in kubelet's `ExecPreStart` and I was able to observe systemd service messages for kubelet after I logged into the admin container:
```
[erikson@admin]$ [FAILED] Failed to start Kubelet.                                                                                 
See 'systemctl status kubelet.service' for details.                                                                                
[  OK  ] Reached target Multi-User System.                                                                                                                                                                                                                            
[erikson@admin]$ sudo sheltie 
```
Not quite sure why it failed after the sleep was done. It automatically restarted and fixed itself though.

etungsten:
Push above fixes some lingering comments.

etungsten:
Push above adds documentation about the new `password-hash` field in the README.

bcressey (approved):
🚀

jpculp (approved):

etungsten:
Push above adds a change to remove surrounding quotes if there are surrounding quotes around the config values:
```
[erikson@admin]$ cat /proc/cmdline 
console="tty0" console="ttyS1,115200n8" ...
[erikson@admin]$   CONSOLES=()
[erikson@admin]$   for opt in $(cat /proc/cmdline) ; do
>      optarg="$(expr "${opt}" : '[^=]*=\(.*\)' ||:)"
>      optarg="${optarg%\"}"
>      optarg="${optarg#\"}"
>      case "${opt}" in
>         console=*) CONSOLES+=("${optarg%,*}") ;;
>      esac
>   done

[erikson@admin]$ echo ${CONSOLES[@]}
tty0 ttyS1
```

jpculp (approved):

etungsten:
Push above fixes an issue where user creation may fail when the host restarts and the user already exists.
Push above also redirects some useless warnings/errors to `/dev/null`

jpculp (approved):
Good catch!

etungsten:
Push above prevents an `jq` invocation from failing the script.
It also adds a change to make the admin container sleep forever if the host-container user-data is unset to prevent errors later on in the script that will make the container fail.

jpculp (approved):

bcressey (approved):

PR #60, opened 2022-06-03
Symlink logger with true

jpculp:
**Description of changes:**

This addresses a regression where logger no longer has an interface to log to with our minimal implementation of systemd. This was causing EC2 Instance Connect to fail to parse keys and make them available on the host.

**Testing done:**

- Launched `aws-ecs-1` ami with new container set in userdata.
- Enabled the admin container via the control container's APIclient.
- Verified APIclient exec functionality.
- Verified Bottlerocket access via `sudo sheltie`.
- Connected to admin container via ec2 instance connect.

**Terms of contribution:**

### DIFF
--- a/start_admin.sh
+++ b/start_admin.sh
@@ -170,7 +170,7 @@ if kex_algorithms=$(jq -r -e -c '.["ssh"]["key-exchange-algorithms"]? | join(","
   echo "KexAlgorithms ${kex_algorithms}" >> "${SSHD_CONFIG_FILE}"
 fi
 
-# Check the configurations are for EC2 instance connect
+# Check the configurations are for EC2 Instance Connect
 declare -i use_eic=0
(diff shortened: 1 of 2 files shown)


start_admin.sh
```diff
@@ -111,6 +111,9 @@ enable_systemd_services() {
   done
   # Enable the SSH daemon service unit so we can run it in the background
   systemctl --user enable "sshd.service"
+  # Reduces issues related to logger and our implementation of systemd. This is
+  # necessary for scripts logging to logger, such as in EC2 Instance Connect.
+  ln -sf /usr/bin/true /usr/bin/logger
```
bcressey: Why not do this in the Dockerfile instead?

bcressey (approved):

bcressey (approved):

jpculp:
Moved symlinking from entrypoint script to Dockerfile.

etungsten (approved):

Issue #75, opened 2023-01-11
admin container should run on hosts with cgroup v2 enabled


markusboehme:
**Image I'm using:** any Bottlerocket variant, any admin container image

**Issue or Feature Request:**

The admin container should be able to run on Bottlerocket hosts that use cgroup v2/the unified cgroup hierarchy. Currently, the admin container cannot start on such hosts and will exit with the following errors:

```
[   15.511906] host-ctr[1189]: Cannot determine cgroup we are running in: No such file or directory
[   15.512878] host-ctr[1189]: Failed to allocate manager object: No such file or directory
[   16.027735] host-ctr[1189]: time="2023-01-11T15:49:06Z" level=fatal msg="Container admin exited with non-zero status"
```

The reason is Amazon Linux 2 in the base image shipping with systemd v219. This version of systemd does not support cgroup v2.

markusboehme:
With the approach in #76 being insufficient for non-superpowered admin containers that cannot mount filesystems on their own, another approach is needed.

Untested idea: Some of the compatibility work spills over into `host-ctr` that mounts a cgroup v1 root in `/sys/fs/cgroup` on demand instead of the default bind mount of the host's `/sys/fs/cgroup`. The functionality could hide behind a flag for `host-ctr` that is passed by `host-containers` exclusively for the admin container.

The elephant in the room is the unattractive special-casing of the admin container. It could be removed once the admin container image moves to a base image that supports cgroup v2. Alternatively, the `host-ctr` flag could generically be exposed via the settings model, but there would be no good choice for a default that fits both now and in a cgroup v2-only future.

markusboehme:
> Untested idea: Some of the compatibility work spills over into host-ctr that mounts a cgroup v1 root in `/sys/fs/cgroup` on demand instead of the default bind mount of the host's `/sys/fs/cgroup`. The functionality could hide behind a flag for `host-ctr` that is passed by `host-containers` exclusively for the admin container.

After hacking a bit on this and looking at code of containerd and runc, this appears to not be a great idea. cgroup filesystems are understandably special-cased in both of them.

markusboehme:
Cleaning up loose ends: #76 was merged (see more discussion there) and released with v0.10.0.

Issue #62, opened 2022-07-05
systemd warning about conflicting jobs when the admin container exits


etungsten:
**Image I'm using:**
[v0.9.0](https://github.com/bottlerocket-os/bottlerocket-admin-container/releases/tag/v0.9.0)

**Issue or Feature Request:**

When exiting the admin container, the user instance of systemd will log the following into the journal which seems concerning:
```
systemd[1]: Stopping Host container: admin...
host-ctr[1648]: Requested transaction contains unmergeable jobs: Transaction contains conflicting jobs 'stop' and 'start' for systemd-exit.service. Probably contradicting requirement dependencies configured.
host-ctr[1648]: Failed to enqueue exit.target job: Transaction contains conflicting jobs 'stop' and 'start' for systemd-exit.service. Probably contradicting requirement dependencies configured.
host-ctr[1648]: time="2022-07-05T17:26:16Z" level=info msg="container task exited" code=0
systemd[1]: host-containers@admin.service: Deactivated successfully.
```

The admin container process still exits non-zero.

markusboehme:
A path I followed for #76 happened to converge with this issue. Here's what I found: Exiting a user instance of systemd starts the `exit.target`, which depends on `systemd-exit.service`, which in turn depends on `-.slice` (the root slice) and `shutdown.target`. The stated conflict is between the latter two, with `-.slice` defining a conflict with `shutdown.target`. This can be fixed via a backport of https://github.com/systemd/systemd/pull/4477 which I'll send a PR for once #76 is merged, since it provides everything needed to build systemd for the admin container.

PR #76, opened 2023-01-11
Run on cgroup v2 host

markusboehme:
**Issue number:** #75 

**Description of changes:** This PR aims to make the admin container work on Bottlerocket hosts with cgroup v2/the unified cgroup hierarchy enabled (rather than the current cgroup v1 default). The current version of the admin container is hindered by systemd v219 inside its image not supporting cgroup v2, and no more recent version of systemd being available for the Amazon Linux 2 base image.

The PR introduces a dedicated but unused cgroup v1 hierarchy just for the admin container. It instructs the user systemd process inside the container to use this separate hierarchy via downstream patches. In order to apply those systemd needs to be recompiled during the build process.

The core of this goes back to @bcressey noting the cgroup v1 filesystem can be mounted without any controllers. I take responsibility for actually running with this and deciding to patch systemd inside the container. The benefit of this compared to doing it on the host is that all code required for the admin container actually stays inside the admin container, and no changes at all can be observed on hosts that are not running the admin container.

**Testing done:** Locally with cgroup v1 and cgroup v2 VMs.

To test with cgroup v2 enabled, add the following user data snippet on boot

```
[settings.boot]
reboot-to-reconcile = true

[settings.boot.init]
"systemd.unified_cgroup_hierarchy" = ["1"]
```

or apply the corresponding settings at runtime using `apiclient set -j '{"settings": {"boot": {"init": {"systemd.unified_cgroup_hierarchy": ["1"]}}}}'` and reboot.

The container image with the changes from this PR can be used both on a cgroup v1 and a cgroup v2 host. Host access using `sudo sheltie` works just the same. The admin container can be stopped/disabled and restarted/reenabled.

I launched a few pods into a Kubernetes cluster with a node that used the cgroup v2 hierarchy. I then verified on the host that the containers were running in the cgroup v2 hierarchy.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,15 @@ RUN test -n "$IMAGE_VERSION"
 LABEL "org.opencontainers.image.version"="$IMAGE_VERSION"
 
 RUN yum update -y \
-    && yum install -y openssh-server sudo shadow-utils util-linux procps-ng jq openssl ec2-instance-connect \
+    && yum install -y
+        ec2-instance-connect \
+        jq \
+        openssh-server \
+        openssl \
+        procps-ng \
+        shadow-utils \
+        sudo \
+        util-linux \
     && yum clean all
 # Delete SELinux config file to prevent relabeling with contexts provided by the container's image
 RUN rm -rf /etc/selinux/config
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,6 +72,8 @@ RUN last_patch=$(awk '/^Patch[0-9]+/ { line = NR } END { print line }' systemd.s
     { \
         echo ;\
         echo '# Bottlerocket Patches'; \
+        echo 'Patch9501: 9500-cgroup-util-extract-cgroup-hierarchy-base-path-into-.patch'; \
+        echo 'Patch9502: 9501-cgroup-util-accept-cgroup-hierarchy-base-as-option.patch'; \
         echo ; \
     } >>systemd.mod.spec; \
     tail -n+$((last_patch + 1)) systemd.spec >>systemd.mod.spec; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,14 @@
-FROM public.ecr.aws/amazonlinux/amazonlinux:2 as builder
+################################################################################
+# Base image for all builds
+
+FROM public.ecr.aws/amazonlinux/amazonlinux:2 as builder-base
 RUN yum group install -y "Development Tools"
+
+
+################################################################################
+# Statically linked, more recent version of bash
+
+FROM builder-base as builder-static
 RUN yum install -y glibc-static
 
 ARG musl_version=1.2.3
@@ -38,6 +47,10 @@ RUN cp bash /opt/bash
 RUN mkdir -p /usr/share/licenses/bash && \
     cp -p COPYING /usr/share/licenses/bash
 
+
+################################################################################
+# Actual admin container image
+
 FROM public.ecr.aws/amazonlinux/amazonlinux:2
 
 ARG IMAGE_VERSION
@@ -51,8 +64,8 @@ RUN yum update -y \
 # Delete SELinux config file to prevent relabeling with contexts provided by the container's image
 RUN rm -rf /etc/selinux/config
 
-COPY --from=builder /opt/bash /opt/bin/
-COPY --from=builder /usr/share/licenses/bash /usr/share/licenses/bash
+COPY --from=builder-static /opt/bash /opt/bin/
+COPY --from=builder-static /usr/share/licenses/bash /usr/share/licenses/bash
 
 RUN rm -f /etc/motd /etc/issue
 COPY --chown=root:root motd /etc/
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@
 
 FROM public.ecr.aws/amazonlinux/amazonlinux:2 as builder-base
 RUN yum group install -y "Development Tools"
+RUN useradd builder
 
 
 ################################################################################
@@ -49,6 +50,36 @@ RUN mkdir -p /usr/share/licenses/bash && \
 
 
 ################################################################################
+# Rebuild of Amazon Linux 2's systemd v219 with downstream patches
+
+FROM builder-base AS builder-systemd
+RUN yum install -y yum-utils rpm-build
+RUN yum-builddep -y systemd
+
+USER builder
+WORKDIR /home/builder
+RUN yumdownloader --source systemd
+RUN rpm -Uv systemd-219-*.src.rpm
+
+WORKDIR /home/builder/rpmbuild/SOURCES
+COPY systemd-patches/*.patch ./
+
+WORKDIR /home/builder/rpmbuild/SPECS
+# Recreate the spec file from three parts: everything up until the last upstream
+# patch, downstream patches, everything else.
+RUN last_patch=$(awk '/^Patch[0-9]+/ { line = NR } END { print line }' systemd.spec); \
+    head -n${last_patch} systemd.spec >systemd.mod.spec; \
+    { \
+        echo ;\
+        echo '# Bottlerocket Patches'; \
+        echo ; \
+    } >>systemd.mod.spec; \
+    tail -n+$((last_patch + 1)) systemd.spec >>systemd.mod.spec; \
+    mv systemd.mod.spec systemd.spec
+RUN rpmbuild --bb systemd.spec
+
+
+################################################################################
 # Actual admin container image
 
 FROM public.ecr.aws/amazonlinux/amazonlinux:2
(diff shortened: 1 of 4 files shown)


markusboehme:
Elaborated on how I tested the change.

markusboehme:
In my built on aarch64, the new container image weighs in at 417 MB vs 324 MB as reported by `docker image ls`. A quick tally over the respective images' root directories comes out roughly equal, so I suspect there might be some whiteouts from upper image layers. Potentially the old files of the original systemd RPMs?

markusboehme:
There is an (experimental) `--squash` option to `docker build` to merge image layers and remove redundant files. This does have the desired effect of squeezing the overall image size to 316 MB (8 MB less than the image before this PR). Naturally this will also prevent sharing of image layers, which at least would be helpful for the Amazon Linux 2 base image being shared between the standard admin and control container images.

Another quick win can be had by bind-mounting the newly built RPMs from a previous build stage rather than temporarily copying them in. This gets the image down to 372 MB (45 MB saved, but still an unnecessary 48 MB heavier than before this PR) without the drawbacks of squashing the layers.

I'll update the PR with the bind-mount approach. Perhaps somebody has another idea to squeeze the image some more.

Dockerfile
```diff
@@ -1,5 +1,14 @@
-FROM public.ecr.aws/amazonlinux/amazonlinux:2 as builder
+################################################################################
+# Base image for all builds
+
+FROM public.ecr.aws/amazonlinux/amazonlinux:2 as builder-base
```
bcressey, line 4: nit: typo in commit message, "sparate build stage"

Dockerfile
```diff
@@ -48,6 +48,35 @@ RUN mkdir -p /usr/share/licenses/bash && \
...
+RUN yum-builddep -y systemd
+
+WORKDIR /root
+RUN yumdownloader --source systemd
+RUN rpm -Uv systemd-219-*.src.rpm
+
+WORKDIR /root/rpmbuild/SOURCES
+COPY systemd-patches/*.patch ./
+
+WORKDIR /root/rpmbuild/SPECS
```
bcressey: nit: I have a bit of a reflexive dislike of building RPMs as root - can we use a throwaway unprivileged user account for this instead?

Dockerfile
```diff
@@ -48,6 +48,35 @@ RUN mkdir -p /usr/share/licenses/bash && \
...
+# Recreate the spec file from three parts: everything up until the last upstream
+# patch, downstream patches, everything else.
+RUN last_patch=$(grep -En '^Patch[0-9]+' systemd.spec | tail -1 | cut -f1 -d:); \
```
bcressey: maybe this instead of the `grep` / `tail` / `cut` pipeline?
```
 awk '/^Patch[0-9]+/ { line=NR } END { print line }' systemd.spec
```

start_admin.sh
```diff
@@ -212,7 +212,34 @@ install_proxy_profile
 
 enable_systemd_services
 
+systemd_options=()
+
+# cgroup v2 compatibility crimes: systemd 219 in the admin container only
+# supports cgroup v1.
+if grep ' /sys/fs/cgroup cgroup2 ' /proc/self/mounts; then
```
bcressey: If we have `findmnt`, this would be more precise:
```
findmnt /sys/fs/cgroup -o FSTYPE -n
```

start_admin.sh
```diff
@@ -212,7 +212,34 @@ install_proxy_profile
...
+    # documentation advises against manually interfering with anything in that
+    # path. Therefore, mount the extra hierarchy elsewhere and tell systemd in
+    # the admin container where to look. Requirements:
+    #
+    #   1. The base must be a mount point.
+    #   2. The base must contain another mounted cgroup file system named "systemd".
+    #
+    # In either case there is no need to populate the file systems with any
+    # actual cgroup controllers, since the host is expected to manage processes.
+    readonly cgroup_base=/bottlerocket-admin/cgroup
```
bcressey: nit: perhaps `/.bottlerocket/cgroup` to "hide" the mount with other host container specific files

start_admin.sh
```diff
@@ -212,7 +212,34 @@ install_proxy_profile
...
+    mkdir -p "${cgroup_base}"
+    mount -t tmpfs -o nosuid,nodev,noexec tmpfs "${cgroup_base}"
+    mkdir -p "${cgroup_base}"/systemd
+    mount -o remount,ro "${cgroup_base}"
+    mount -t cgroup -o none,name=systemd cgroup "${cgroup_base}"/systemd
```
bcressey, line 238: When this mount is added, all processes on the system will have it appear in `/proc/<PID>/cgroup`. Does the entry go away if the admin container is subsequently disabled and this is unmounted?

bcressey (approved):
Nice work. It's an ugly thing to have to do, but you found a clean way to do it.

bcressey:
> I'll update the PR with the bind-mount approach. Perhaps somebody has another idea to squeeze the image some more.

systemd isn't in the base container image; it comes in from when the "Development Tools" group is installed. If the layers are structured so that we don't install that group until the end - or potentially not at all, if we replace it with a more targeted package install command - then we could avoid the whiteout of the previous systemd install.

bcressey:
Oh, there's already a more targeted `yum install` - I think we just need to install the RPMs from the bind mount before rather than after that point.

Dockerfile
markusboehme: Good point. Added a separate unprivileged `builder` user for those steps that don't need to run as root.

Dockerfile
markusboehme: I love a good awk one-liner, and this one makes it much clearer what's happening. Thanks!

start_admin.sh
markusboehme: Good call-out, I had forgotten about `/.bottlerocket`. I chose the more verbose `/.bottlerocket/admin-container-cgroup` as the root.

start_admin.sh
markusboehme: Now this is a bit peculiar. It actually does not, and does not match my expectations (e.g. compare with performing the equivalent steps via `sudo unshare -m`). I still need to dig into this some more. It seems to relate to how the admin container is disabled (e.g. if systemd exits voluntarily before entering its main loop, the mount is cleaned up). So far I could not find any process remaining in the admin container mount namespace and don't understand why the cgroup root seems to remain.

markusboehme:
>> I'll update the PR with the bind-mount approach. Perhaps somebody has another idea to squeeze the image some more.
>
> systemd isn't in the base container image; it comes in from when the "Development Tools" group is installed. If the layers are structured so that we don't install that group until the end - or potentially not at all, if we replace it with a more targeted package install command - then we could avoid the whiteout of the previous systemd install.

True, I jumped to conclusions here, thanks for noticing! systemd is not in the base image, but actually pulled in as a dependency of the `openssh-server` package. This means we can cut the dead weight by making sure the downstream systemd build is installed alongside the other packages baked into the admin container image.

markusboehme:
The earlier force push addresses feedback by @bcressey. Leaving the PR in draft mode until I figure out why the cgroup root lingers on when the admin container is disabled.

start_admin.sh
markusboehme: I understand this better now. When systemd goes to remove the cgroup of the `sshd.service`, its call to `unlinkat` returns `EBUSY`. The cgroup is never removed. The cgroup root lingers on as long as there's still a cgroup in it, even if it's empty. Even if there doesn't seem to be a way to get back into the mount namespace. Aside from figuring out why removing the cgroup fails, this could also be cleaned up later by remounting the cgroup root with the same name--it'll get reused--and reaping empty cgroups.

While looking at this I found something I had completely overlooked so far, however, and I'll likely have to scrap the entire implementation. :disappointed: See my comment in the main discussion below.

markusboehme:
What I had completely overlooked so far is that the admin container may be run without superpowers (`superpowered = false` in the user data). It then will not have the necessary permissions to mount a cgroup v1 root by itself in `start_admin.sh`. While there's other problems in running without superpowers at the moment (#78), this probably tanks the approach of containing all the compatibility work in the admin container.

jpculp (approved):

start_admin.sh
markusboehme: An update on the lingering cgroup root. There are two issues at play here:

1. The empty cgroup for `sshd.service` not being removed on exit. The reason for this turns out to be #62. When systemd fails to start `exit.target`, it makes a dash for the exit. Fixing this issue leads to a proper exit and all cgroups being removed.
1. A much more interesting timing-dependent kernel behavior. I'm dedicating the rest of the comment to this one.

Unmounting a cgroup file system can trigger removal of a cgroup root if it was the last mount of that root and the root is otherwise unused. Since with 1. out of the way all cgroups are removed before the container exits, leaving its mount namespace should trigger removal of the cgroup root. However, since the removal of an individual cgroup is an asynchronous multi-step endeavor, unmount and cgroup removal can actually race, with the unmount happening before a cgroup has been fully removed. Once the `rmdir` syscall returns, user-visible effects of cgroup removal have occured, but internal bookkeeping tasks are not yet complete and happen asynchronously. In this case, the cgroup root lingers on and can be observed in e.g. `/proc/1/cgroup` even after the admin container exited.

The issue can be trivially reproduced with e.g. the following sequence of commands (note that `rmdir` and `umount` will need to happen in short succession, so it's best to run this as a script):

```
mkdir -p /mnt/cgroot
mount -t cgroup -o none,name=cgroot cgroup /mnt/cgroot
mkdir /mnt/cgroot/cg
rmdir /mnt/cgroot/cg
umount /mnt/cgroot
```

I spent some time looking at fixing this from the kernel side, but the highly asynchronous nature of the cgroup release code makes this a delicate matter with plenty opportunity to get things wrong. Knowing what to look for now, I noticed I haven't been the first one to observe or try to fix this one either. [The discussion on the cgroups mailing list](https://lore.kernel.org/lkml/20220404142535.145975-1-minhquangbui99@gmail.com/) concluded not to fix this race.

For Bottlerocket's purposes, this is a cosmetic issue. Nothing is really being leaked. The same cgroup root is available and gets reused when the admin container is started again. I'd therefore propose to just live with that for the time being (a workaround would be to re-mount/unmount the hierarchy after some time, e.g. in a `host-ctr` clean-up hook).

markusboehme:
I will merge this tomorrow unless someone speaks up against it. Consensus on #78 is to not worry about the "admin container without superpowers" use case, since it's been broken for a while already without anyone noticing and it not being super useful if it weren't.

bcressey (approved):
Nicely done!

Issue #84, opened 2023-08-24
Stop unnecessary reverse lookup of incoming SSH connections


markusboehme:
**Image I'm using:** Any Bottlerocket image with the latest admin container v0.10.2

**Issue or Feature Request:** When logging into the admin container via SSH, a multitude of DNS requests are sent. Here's a subset of them observed via `tcpdump 'port 53'` inside the admin container in a local virtualized environment (10.0.2.15 is the virtual interface, 10.0.2.2 the virtual gateway, 10.0.2.3 QEMU's virtual resolver):

```
20:43:19.268506 IP ip-10-0-2-15.eu-central-1.compute.internal.60800 > ip-10-0-2-3.eu-central-1.compute.internal.domain: 49101+ PTR? 2.2.0.10.in-addr.arpa. (39)
20:43:19.270473 IP ip-10-0-2-3.eu-central-1.compute.internal.domain > ip-10-0-2-15.eu-central-1.compute.internal.60800: 49101 1/0/0 PTR ip-10-0-2-2.eu-central-1.compute.internal. (94)
20:43:19.270691 IP ip-10-0-2-15.eu-central-1.compute.internal.42991 > ip-10-0-2-3.eu-central-1.compute.internal.domain: 16137+ A? ip-10-0-2-2.eu-central-1.compute.internal. (59)
20:43:19.271989 IP ip-10-0-2-3.eu-central-1.compute.internal.domain > ip-10-0-2-15.eu-central-1.compute.internal.42991: 16137 1/0/0 A 10.0.2.2 (75)
20:43:19.291881 IP ip-10-0-2-15.eu-central-1.compute.internal.45672 > ip-10-0-2-3.eu-central-1.compute.internal.domain: 45883+ A? ip-10-0-2-2.eu-central-1.compute.internal. (59)
20:43:19.291930 IP ip-10-0-2-15.eu-central-1.compute.internal.45672 > ip-10-0-2-3.eu-central-1.compute.internal.domain: 33852+ AAAA? ip-10-0-2-2.eu-central-1.compute.internal. (59)
20:43:19.293347 IP ip-10-0-2-3.eu-central-1.compute.internal.domain > ip-10-0-2-15.eu-central-1.compute.internal.45672: 45883 1/0/0 A 10.0.2.2 (75)
[...]

PR #85, opened 2023-08-24
Stop unnecessary reverse lookup of incoming SSH connections

markusboehme:
**Issue number:** #84 

**Description of changes:** Disable reverse lookup of hostnames for incoming SSH connections. Neither the sshd configuration nor the authorized_keys file generated at the start of the admin container make use of the hostname information, so save the unnecessary requests.

**Testing done:** Can still connect to the admin container both via explicit key pair and via EC2 Instance Connect on both cgroup v1 (aws-k8s-1.23) and cgroup v2 (aws-k8s-1.27) hosts. Observing `tcpdump 'port 53'` from within the admin container shows no more DNS requests on incoming SSH connections.

**Terms of contribution:**

### DIFF
--- a/sshd_config
+++ b/sshd_config
@@ -19,3 +19,5 @@ PermitRootLogin no
 
 # Configured by user data
 TrustedUserCAKeys /etc/ssh/trusted_user_ca_keys.pub
+
+UseDNS no


cbgbt (approved):
This seems good to me, but I wonder if the looked-up-names are logged and possibly useful in a forensics situation.

bcressey (approved):

yeazelm (approved):

stmcginnis (approved):

markusboehme:
> This seems good to me, but I wonder if the looked-up-names are logged and possibly useful in a forensics situation.

Yes, this is worth calling out. With this change there's no way to log the looked-up host name (supposedly) belonging to an incoming connection. Since the configured sshd will not take any decisions on the host name it does not seem worth preserving to me, however. sshd will still keep track of actual established sessions in `/var/log/wtmp` and the host name listed there will be the remote IP. If needed the reverse lookup may be performed using that information at a later time.

Issue #90, opened 2023-12-14
Feature Request: Add settings for SSHD MACs


mlacko64:
**Issue or Feature Request:**

Good day,
Is it possible to add support for setting MACs in sshd config? Currently ciphers and key-exchange-algorithms can be changed, but our vulnerability scanner is also reporting that there are used deprecated MACs, which we cannot change now in admin container.

Thank you very much

mlacko64:
Created a pull request https://github.com/bottlerocket-os/bottlerocket-admin-container/pull/91 with code change

markusboehme:
Hi @mlacko64, thanks again for your PR! The feature has been released with admin container version v0.11.3 as used in Bottlerocket v1.18.0 and later.

PR #91, opened 2023-12-15
Allow setting MACs in SSHD config

mlacko64:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket-admin-container/issues/90

**Description of changes:**

This change adds option to customize MACs for SSH , for example, to disable SHA1 MACs which are reported as deprecated by vulnerability scanner. README.md updated.

**Testing done:**

Created custom container and tested in my lab cluster, works as expected.

SSHD config contains line:
`MACs hmac-sha2-256,hmac-sha2-512,hmac-md5,hmac-md5-96,umac-64@openssh.com,umac-128@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com,hmac-md5-etm@openssh.com,hmac-md5-96-etm@openssh.com,umac-64-etm@openssh.com,umac-128-etm@openssh.com`

SSHD server offers just MACs selected by me:
```
debug2: MACs ctos: hmac-sha2-256,hmac-sha2-512,hmac-md5,hmac-md5-96,umac-64@openssh.com,umac-128@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com,hmac-md5-etm@openssh.com,hmac-md5-96-etm@openssh.com,umac-64-etm@openssh.com,umac-128-etm@openssh.com
debug2: MACs stoc: hmac-sha2-256,hmac-sha2-512,hmac-md5,hmac-md5-96,umac-64@openssh.com,umac-128@openssh.com,hmac-sha2-256-etm@openssh.com,hmac-sha2-512-etm@openssh.com,hmac-md5-etm@openssh.com,hmac-md5-96-etm@openssh.com,umac-64-etm@openssh.com,umac-128-etm@openssh.com
```
**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,31 @@ To change allowed key exchange algorithms to a specific set, you can add a
 }
 ```
 
+To change allowed MACs to a specific set, you can add a `macs` section:
+
+```json
+{
+  "ssh": {
+    "authorized-keys...",
+    "macs": [
+      "hmac-sha2-256",
+      "hmac-sha2-512",
+      "umac-64@openssh.com",
+      "umac-128@openssh.com",
+      "hmac-sha2-256-etm@openssh.com",
+      "hmac-sha2-512-etm@openssh.com",
+      "hmac-md5-etm@openssh.com",
+      "hmac-md5-96-etm@openssh.com",
+      "umac-64-etm@openssh.com",
+      "umac-128-etm@openssh.com"
+    ]
+  }
+}
+```
+
+You can also tweak ciphers, key exchange algorithms and MACs following way (see https://man.openbsd.org/sshd_config for details):
+- If the specified list begins with a ‘+’ character, then the specified entries will be appended to the default set instead of replacing them. If the specified list begins with a ‘-’ character, then the specified entries (including wildcards) will be removed from the default set instead of replacing them. If the specified list begins with a ‘^’ character, then the specified entries will be placed at the head of the default set.
+
 By default, the admin container's local user will be `ec2-user`. If you would like to change this, you can set the user value like so:
 
 ```json
--- a/start_admin.sh
+++ b/start_admin.sh
@@ -170,6 +170,11 @@ if kex_algorithms=$(jq -r -e -c '.["ssh"]["key-exchange-algorithms"]? | join(","
   echo "KexAlgorithms ${kex_algorithms}" >> "${SSHD_CONFIG_FILE}"
 fi
 
+# Populate MACs with all the MACs found in user-data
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -84,6 +84,30 @@ To change allowed key exchange algorithms to a specific set, you can add a
...
+
+```json
+{
+  "ssh": {
+    "authorized-keys...",
+    "macs": [
+      "hmac-sha2-256",
+      "hmac-sha2-512",
+      "hmac-md5",
+      "hmac-md5-96",
```
max-boehm: maybe drop weak md5 hmacs from the example?

max-boehm (approved):

webern (commented):
Thank you for this.

I'm not a subject matter expert, but this makes sense to me. We allow the configuration of Ciphers and KexAlgorithms, so it feels like MACs would be intended to be configurable as well.

Can you squash these into one commit? 4b2afc6 9b9c2b3

README.md
jahkeup: Good call out. And you know, rather than give an explicit list example maybe show how one might use the setting to tweak the default? Is that intended to be supported here @mlacko64 ?

ie: if we consider https://man.openbsd.org/sshd_config#MACs

> If the specified list begins with a ‘+’ character, then the specified algorithms will be appended to the default set instead of replacing them. If the specified list begins with a ‘-’ character, then the specified algorithms (including wildcards) will be removed from the default set instead of replacing them. If the specified list begins with a ‘^’ character, then the specified algorithms will be placed at the head of the default set.

I think we'd be able to express surgical "edits" to the defaults - maybe even this would be a model default setting? Something like this:

```json
{
  "ssh": {
    "macs": [
      "-hmac-sha1"
    ]
  }
}
```

This way, any Bottlerocket settings (ie: other than build time defaults of OpenSSH itself) would be able to augment rather than exhaustively set the list. Settings could even be provided in the model to default _remove_ insecure elements if/when the need arises.

In fact, after chatting with @jpculp I think this might apply to a few of the other "list" settings in `ssh`.

mlacko64:
Thanks for comments, I have:
- removed "hmac-md5" and "hmac-md5-96" from example
- updated readme.md with info that ciphers, key exchange algorithms and MACs can be also just added or removed from list
- squashed all commits into one

webern (approved):

jpculp:
@mlacko64, thanks for doing this! I think we can probably go ahead and merge this into an upcoming container release, but would you be willing to prune the commit message beforehand? The top line is probably enough to suffice.

mlacko64:
@jpculp sure, I have pruned commit message as you suggested

jpculp (approved):

Issue #112, opened 2025-02-28
Possibility to change default SSH port


dpavlov-smartling:
**Image I'm using:**
bottlerocket-admin:v0.11.13

**Issue or Feature Request:**
Due to security reasons we usually change default SSH port 22 on some alternative port. As far as I can see currently you can't change SSH port for the admin container.
Do you think this can be implemented or are you interested in such PR?

Thanks,

jpculp:
Hi @dpavlov-smartling, thank you for reaching out! Apologies for the delayed response. I think this is something that can be supported relatively easily using the same logic we use to customize other settings such as the allowed ciphers. I can add this to our backlog and assign to me (at least for now) so it doesn't get lost. Of course if you're interested in contributing the feature we'd be more than happy to review.

maherthomsi:
Hi @dpavlov-smartling, the feature you requested has been released with admin container [v0.12.0](https://github.com/bottlerocket-os/bottlerocket-admin-container/releases/tag/v0.12.0).

PR #118, opened 2025-06-04
Allow setting port in SSHD config

maherthomsi:
**Issue number:**

#112

**Description of changes:**

Add support for customizing the SSH daemon port in the admin container. README.md updated.

**Testing done:**
Built and deployed a custom admin container with the port customization feature.
Verified functionality by configuring a non-standard port (1234) in user-data, then confirmed the SSH daemon was correctly listening on the specified port using `ss -tlnp | grep sshd`
Admin container connected and works as expected.

**Terms of contribution:**

### DIFF
--- a/start_admin.sh
+++ b/start_admin.sh
@@ -185,6 +185,11 @@ if macs=$(jq -r -e -c '.["ssh"]["macs"]? | join(",")' "${USER_DATA}" 2>/dev/null
   echo "MACs ${macs}" >> "${SSHD_CONFIG_FILE}"
 fi
 
+# Set custom SSH port if specified in user-data
+if ssh_port=$(jq -r -e '.["ssh"]["port"]?' "${USER_DATA}" 2>/dev/null); then
+  echo "Port ${ssh_port}" >> "${SSHD_CONFIG_FILE}"
+fi
+
 # Check the configurations are for EC2 Instance Connect
 declare -i use_eic=0
 if [[ $authorized_keys_command == /opt/aws/bin/eic_run_authorized_keys* ]] \
--- a/README.md
+++ b/README.md
@@ -106,6 +106,17 @@ To change allowed MACs to a specific set, you can add a `macs` section:
 }
 ```
 
+To change the port that the admin container SSH daemon is running on (default is 22), you can add a `port` section:
+
+```json
+{
+  "ssh": {
+    "authorized-keys...",
+    "port": 1234
+  }
+}
+```
+
 You can also tweak ciphers, key exchange algorithms and MACs following way (see https://man.openbsd.org/sshd_config for details):
 - If the specified list begins with a ‘+’ character, then the specified entries will be appended to the default set instead of replacing them. If the specified list begins with a ‘-’ character, then the specified entries (including wildcards) will be removed from the default set instead of replacing them. If the specified list begins with a ‘^’ character, then the specified entries will be placed at the head of the default set.
 


start_admin.sh
```diff
@@ -181,8 +181,19 @@ if kex_algorithms=$(jq -r -e -c '.["ssh"]["key-exchange-algorithms"]? | join(","
...
-if macs=$(jq -r -e -c '.["ssh"]["macs"]? | join(",")' "${USER_DATA}" 2>/dev/null); then
-  echo "MACs ${macs}" >> "${SSHD_CONFIG_FILE}"
+  if macs=$(jq -r -e -c '.["ssh"]["macs"]? | join(",")' "${USER_DATA}" 2>/dev/null); then
+    echo "MACs ${macs}" >> "${SSHD_CONFIG_FILE}"
+  fi
+
+# Set custom SSH port if specified in user-data
+if ssh_port=$(jq -r -e '.["ssh"]["port"]?' "${USER_DATA}" 2>/dev/null); then
+  # Validate port number (must be between 1 and 65535)
+  if [[ "${ssh_port}" =~ ^[0-9]+$ ]] && [ "${ssh_port}" -ge 1 ] && [ "${ssh_port}" -le 65535 ]; then
```
jpculp:
```suggestion
  if [[ ${ssh_port} -ge 1 ]] && [[ ${ssh_port} -le 65535 ]]; then
```
This can be simplified a bit. Also no need to quote ints. The ssh server will do its own validation so you could even just do `[[ ${ssh_port} -ge 1 ]]`

start_admin.sh
```diff
@@ -181,8 +181,19 @@ if kex_algorithms=$(jq -r -e -c '.["ssh"]["key-exchange-algorithms"]? | join(","
...
+    echo "Port ${ssh_port}" >> "${SSHD_CONFIG_FILE}"
+    log "Setting custom SSH port: ${ssh_port}"
+  else
+    log "Invalid SSH port specified: ${ssh_port}. Using default port = 22."
```
jpculp: Lets not do this. If someone is attempting to switch the port to something other than 22 we shouldn't be falling back to the port they were trying to avoid. Better to rely on the ssh server failing and returning its own error.

jpculp (changes requested):

start_admin.sh
maherthomsi: That makes sense, would I still run the previous conditional in that case (port range conditionals) or would I just always write whatever is provided in the user-data and rely on the ssh server to catch it and fail?

jpculp: Probably just write whatever (consistent with the code blocks above), unless we don't think the ssh server is going to handle the error itself.

start_admin.sh
maherthomsi: Removed conditionals in revision

start_admin.sh
maherthomsi: Removed conditionals in revision

maherthomsi (commented):
Remove conditional statements when checking user-data

maherthomsi:
When testing garbage value as port input in user-data.toml (ex. Port: "twenty-two"), these were the SSHD output logs:

```
host-containers@admin[2765]: Setting custom SSH port: twenty-two
host-containers@admin[2765]: sshd.service: main process exited, code=exited, status=255/n/a
host-containers@admin[2765]: Unit sshd.service entered failed state.
host-containers@admin[2765]: sshd.service failed.
```

jpculp (approved):

gthao313 (approved):

